PR #4, opened 2025-02-20
Use non wasm encryption

altkdf:


randombit:
I think this was closed when I deleted the `js-vetkd-utils` branch but it should be rebased on master now?

PR #30, opened 2025-03-20
Add linting using eslint to some of the TS packages

randombit:
Add linting configuration and fix various lints


altkdf (approved):

sdk/ic_vetkd_sdk_encrypted_maps/package.json
```diff
@@ -17,17 +17,19 @@
...
     "@types/jest": "^29.5.14",
     "ts-jest": "^29.2.5",
-    "ts-node": "10.9.2"
+    "ts-node": "10.9.2",
+    "typescript-eslint": "8.24"
   },
   "scripts": {
     "build": "tsc",
-    "lint": "tslint --project tsconfig.json -t codeFrame 'src/**/*.ts' 'test/**/*.ts'",
+    "lint": "eslint",
```
fspreiss, line 30: Should `eslint` also be an explicit dev-dependency (like also shown in the https://typescript-eslint.io/getting-started)?

Same question for the other package.json files.

fspreiss (approved):

sdk/ic_vetkd_sdk_encrypted_maps/package.json
randombit: No idea personally. ChatGPT think we just should just use `eslint`

randombit: Added `eslint` dep to all packages.

PR #33, opened 2025-03-20
test: CRP-2674 add CDK encrypted maps tests

altkdf:


cdk/encrypted_maps/src/lib.rs
```diff
@@ -122,10 +122,7 @@ impl EncryptedMaps {
         caller: Principal,
         key_id: KeyId,
     ) -> Result<Vec<(MapKey, EncryptedMapValue)>, String> {
-        match self.key_manager.get_user_rights(caller, key_id, caller)? {
-            Some(_) => Ok(()),
-            None => Err("unauthorized user".to_string()),
-        }?;
+        self.key_manager.get_user_rights(caller, key_id, caller)?;
```
fspreiss, line 125: Do I understand correctly that what this line does is to ensure that `caller` is authorized to access the map with id `key_id`? If so, then the method name `get_user_rights` is somewhat misleading, because it has the "side effect" that it fails if `caller` is unauthorized.

I'd have expected that a method with the name `get_user_rights` simply returns the user rights, and nothing more.

Not something for this PR, but maybe something to improve in the future.

cdk/encrypted_maps/src/lib.rs
```diff
@@ -142,12 +139,9 @@ impl EncryptedMaps {
...
     ) -> Result<Option<EncryptedMapValue>, String> {
-        match self.key_manager.get_user_rights(caller, key_id, caller)? {
-            Some(_) => Ok(()),
-            None => Err("unauthorized user".to_string()),
-        }?;
-
-        Ok(self.mapkey_vals.get(&(key_id, key)))
+        self.key_manager
+            .get_user_rights(caller, key_id, caller)
+            .map(|_| self.mapkey_vals.get(&(key_id, key)))
```
fspreiss, line 144: Here, this above-mentioned somewhat unexpected side-effect of `get_user_rights` (namely that it also checks for access) is used again. This is also shown by the fact that the result is thrown away.

Also, do I understand correctly that in this `get_encrypted_value` we cannot distinguish between the case where the caller doesn't have access and the case where ther caller does have access but there is no entry for `key` in the map?

If so, it might make sense to consider improving this.

fspreiss (approved):

cdk/encrypted_maps/src/lib.rs
altkdf: > Also, do I understand correctly that in this get_encrypted_value we cannot distinguish between the case where the caller doesn't have access and the case where ther caller does have access but there is no entry for key in the map?

That's not correct. That would return an `Ok(None)`.
```rust
self.key_manager
    // returns Ok(AccessRights) if authorized
    .get_user_rights(caller, key_id, caller)
    // returns Ok(None) if authorized and no `key`
    .map(|_| self.mapkey_vals.get(&(key_id, key))).   
```

cdk/encrypted_maps/src/lib.rs
altkdf: It might be easiest to discuss this offline first.

altkdf: Created [CRP-2739](https://dfinity.atlassian.net/browse/CRP-2739).

PR #42, opened 2025-03-21
chore(crypto): CRP-2721 Add TS linting to CI

randombit:
Linting configurations were added in #31 but the linting step did not run in CI. This change now executes eslint during the CI builds.


sdk/ic_vetkd_sdk_encrypted_maps/src/index.ts
```diff
@@ -24,6 +24,7 @@ export class EncryptedMaps {
                 const mapKey = new TextDecoder().decode(Uint8Array.from(mapKeyBytes.inner));
                 return this.decrypt_for(mapId[0], mapName, mapKey, Uint8Array.from(encryptedValue.inner));
             });
+            /* eslint-disable @typescript-eslint/no-unused-expressions */
```
randombit, line 27: @altkdf I did not really understand what is happening here; is there something that should be fixed/changed? For the moment I've just silenced this lint

randombit:
> Would it make sense to integrate the npm run lints into the respective workflows of the individual packages, rather than having one linting workflow?

Maybe? I guess if nothing else it does seem sound that encrypted_maps and key_manager SDKs should have a CI of some kind

sdk/ic_vetkd_sdk_encrypted_maps/src/index.ts
altkdf: That function is implemented incorrectly and there were neither tests nor places that use this code. I will open a PR once you merge this one.

fspreiss (approved):

PR #47, opened 2025-03-24
Simplify CI provisioning step

randombit:
The CI was provisioning a number of unused/unnecessary dependencies. This change reduces the runtime of some of the CI jobs by several minutes each.


fspreiss (approved):
Yesterday I noticed that node.js waited for a minute on purpose because we used an outdated version, so this PR is great, thanks!

PR #48, opened 2025-03-25
fix: CRP-2745 fixes `get_all_accessible_values()` and adds some tests

altkdf:
Tests:
* get all accessible values
* get all values for a map
* get/set user rights


altkdf:
Context of the fix: https://github.com/dfinity/vetkd-devkit/pull/42#discussion_r2008135337.

sdk/ic_vetkd_sdk_encrypted_maps_example/src/index.test.ts
```diff
@@ -190,20 +190,180 @@ test('get-set roundtrip should be consistent', async () => {
...
+    key4,
+    data4
+  );
+
+  encryptedMapsSharesWithOwner.set_user_rights(
+    sharesWithOwner,
+    mapName1,
+    owner,
+    { Read: null }
+  );
```
fspreiss, line 290: Do I understand correctly that this call has no effect, as long as no map with name `mapName1` exists for owner `sharesWithOwner`? What if `sharesWithOwner` then creates such a map: will `owner` immediately have access then?

Additionally, would it make sense to also add tests that `get_all_accessible_values` works "across" users?

altkdf: This code was a mistake in a few aspects:
* there should be an `await`
* `mapName1` should be `mapName2`
* the test code should check that we are getting 2 maps with 2 keys each

fixed all of that

> Additionally, would it make sense to also add tests that get_all_accessible_values works "across" users?

this should have  been addressed by the fix

altkdf: > Do I understand correctly that this call has no effect, as long as no map with name mapName1 exists for owner sharesWithOwner? What if sharesWithOwner then creates such a map: will owner immediately have access then?

Without an `await`, it never runs :pensive:

sdk/ic_vetkd_sdk_encrypted_maps_example/src/index.test.ts
```diff
@@ -144,20 +146,189 @@ test('get-set roundtrip should be consistent', async () => {
...
+
+  const retrievedValues =
+    await encryptedMapsOwner.get_all_accessible_values();
+
+  // 2 maps
+  expect(retrievedValues.length).to.equal(2);
+  // 2 keys in the first map
+  expect(retrievedValues[0][1].length).to.equal(2);
+  // 2 maps in the second map
```
fspreiss, line 256: typo? 2 maps => 2 keys

sdk/ic_vetkd_sdk_encrypted_maps_example/src/index.test.ts
```diff
@@ -144,20 +146,189 @@ test('get-set roundtrip should be consistent', async () => {
...
+      ownerPrincipal.compareTo(owner) === "eq" &&
+      isEqualArray(
+        Uint8Array.from(mapName.inner),
+        new TextEncoder().encode(mapName1)
+      )
+    ) {
+      const expectedValues: Array<Array<Uint8Array>> = [
+        [new TextEncoder().encode(key1), data1],
+        [new TextEncoder().encode(key2), data2],
+      ];
```
fspreiss, line 275: Shouldn't `owner` now see 4 values? His own two, and the two that `sharesWithOwner` shares with him?

sdk/ic_vetkd_sdk_encrypted_maps_example/src/index.test.ts
altkdf: yes, fixed

sdk/ic_vetkd_sdk_encrypted_maps_example/src/index.test.ts
altkdf: It does. The function returns a vector of `((owner_principal, map_name), Vec<KeyValues>)`, which contains something like
```
[
    (
    [owner, "some map 1"],
        [
            ("some key 1", "Hello, world 1!"),
            ("some key 2", "Hello, world 2!")
        ]
    ),
    (
    [sharesWithOwner, "some map 2"],
        [
            ("some key 3", "Hello, world 3!"),
            ("some key 4", "Hello, world 4!")
        ]
    ),
]
```

fspreiss: Ah, I wrongly read the test as "all values accessible to `owner` are key1/data1 and key2/data; and all values accessible to `sharesWithOwner` are key3/data3 and key4/data4". But the test only retrieves the accessible values once for the owner and the result contains all 4 entries, so all good. Sorry for the noise.

fspreiss (approved):

PR #49, opened 2025-03-25
chore: CRP-2742 add prettier and Svelte linting

altkdf:
* Adds linting workflows to both password manager examples.
* Adds prettier capabilities to all packages.
* Adds formats password manager with metadata.
* Fixes Svelte lints in both password manager examples.


sdk/ic_vetkd_sdk_encrypted_maps/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+    "plugins": ["prettier"],
+    "tabWidth": 4
+}
```
randombit, line 4: Could/should these be shared among packages with a symlink?

altkdf: I guess that was a conscious decision to not have global configs
> Prettier intentionally doesn’t support any kind of global configuration. This is to make sure that when a project is copied to another computer, Prettier’s behavior stays the same.

We could think about it. I doubt though that at the current scale file duplication would become a problem.

altkdf:
This work will be continued in another PR because this one is already too big.

sdk/ic_vetkd_sdk_key_manager_example/package.json
```diff
@@ -16,6 +16,7 @@
     "@dfinity/candid": "^2.3.0",
     "@dfinity/principal": "^2.3.0",
     "@types/node": "^22.13.0",
+    "prettier": "^3.5.3",
```
randombit, line 19: Shouldn't this be a dev dependency?

sdk/ic_vetkd_sdk_encrypted_maps/package.json
```diff
@@ -13,16 +13,20 @@
   },
   "dependencies": {
     "@dfinity/principal": "^2.3.0",
-    "idb-keyval": "^6.2.1"
+    "idb-keyval": "^6.2.1",
+    "prettier": "^3.5.3"
```
randombit, line 17: Should be a dev dependency, right?

altkdf: yes, changed

sdk/ic_vetkd_sdk_utils/package.json
```diff
@@ -8,21 +8,25 @@
     "typings": "dist/types/index.d.ts",
     "dependencies": {
         "@noble/curves": "^1.8.1",
-        "@noble/hashes": "^1.7.1"
+        "@noble/hashes": "^1.7.1",
+        "prettier": "^3.5.3"
```
randombit, line 12: devdep

sdk/ic_vetkd_sdk_key_manager/package.json
```diff
@@ -13,16 +13,20 @@
     },
     "license": "Apache-2.0",
     "dependencies": {
-        "@dfinity/principal": "^2.3.0"
+        "@dfinity/principal": "^2.3.0",
+        "prettier": "^3.5.3"
```
randombit, line 17: devdep

sdk/ic_vetkd_sdk_encrypted_maps_example/package.json
```diff
@@ -15,6 +15,7 @@
         "@dfinity/agent": "^2.3.0",
         "@dfinity/candid": "^2.3.0",
         "@dfinity/principal": "^2.3.0",
+        "prettier": "^3.5.3",
```
randombit, line 18: devdep

sdk/ic_vetkd_sdk_utils/package.json
altkdf: done

sdk/ic_vetkd_sdk_key_manager/package.json
altkdf: done

sdk/ic_vetkd_sdk_encrypted_maps_example/package.json
altkdf: done

sdk/ic_vetkd_sdk_encrypted_maps/.prettierrc
altkdf: I think with the likely future structure of this repo, that we discussed offline, this comment should not be very relevant anymore

altkdf:
This should now be ready for another look, @randombit.

sdk/ic_vetkd_sdk_encrypted_maps/.prettierrc
randombit: Fair

PR #52, opened 2025-03-25
feat: CRP-2746 use javascript utils in encrypted maps and fix its dependents

altkdf:
Compiled artifacts using WASM:
```
dist/index.html                                      0.48 kB │ gzip:   0.30 kB
dist/assets/ic_vetkd_cdk_utils_bg-DvSDd4rZ.wasm    125.17 kB
dist/bundle.css                                     68.02 kB │ gzip:  11.78 kB
dist/assets/index-CbWjwMQl.js                    1,253.67 kB │ gzip: 288.53 kB
```

Compiled artifacts using JS:
```
dist/index.html                    0.48 kB │ gzip:   0.30 kB
dist/bundle.css                   68.02 kB │ gzip:  11.78 kB
dist/assets/index-C3wHOvXs.js  1,053.39 kB │ gzip: 275.94 kB
```

### DIFF
--- a/sdk/ic_vetkd_sdk_encrypted_maps/src/index.ts
+++ b/sdk/ic_vetkd_sdk_encrypted_maps/src/index.ts
@@ -68,7 +68,6 @@ export class EncryptedMaps {
         for (const [x, y] of encryptedValues.Ok) {
             resultGet.push([new TextDecoder().decode(Uint8Array.from(x.inner)), Uint8Array.from(y.inner)]);
         }
-        // console.info("encryptedMaps.get_values_for_map(" + map_owner.toText() + ", " + map_name + " result: " + resultGet);
 
         const result = new Array<[ByteBuf, ByteBuf]>();
         for (const [mapKey, mapValue] of encryptedValues.Ok) {
@@ -80,7 +79,6 @@ export class EncryptedMaps {
         for (const [x, y] of result) {
             resultDecrypted.push([new TextDecoder().decode(Uint8Array.from(x.inner)), new TextDecoder().decode(Uint8Array.from(y.inner))]);
         }
-        // console.info("decrypted encryptedMaps.get_values_for_map(" + map_owner.toText() + ", " + map_name + " result: " + resultDecrypted);
         return { "Ok": result };
     }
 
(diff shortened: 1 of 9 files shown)


altkdf:
Proper JS naming and the full removal of `Result`s in the encrypted maps' API will be addressed more in next PRs.

sdk/ic_vetkd_sdk_encrypted_maps/src/index.ts
```diff
@@ -217,36 +194,36 @@ export interface MapData {
...
-            name: "AES-GCM",
-            iv: iv
-        },
-        key,
-        ciphertext
-    );
-
-    return new Uint8Array(decrypted_bytes);
-}
+// export async function encrypt(bytes_to_encrypt: Uint8Array, key: CryptoKey): Promise<Uint8Array> {
```
randombit, line 197: Can these just be removed now?

sdk/ic_vetkd_sdk_encrypted_maps/src/index.ts
```diff
@@ -176,38 +157,34 @@ export class EncryptedMaps {
...
-            info: new TextEncoder().encode("ic_vetkd_sdk_encrypted_maps_subkey"),
-            hash: "SHA-256",
-        };
-        const derived_key = await window.crypto.subtle.deriveKey(algorithm, get_vetkey_result.Ok, { name: "AES-GCM", length: 256 }, true, ["encrypt", "decrypt"]);
-
-        await set([[map_owner.toString(), map_name, map_key]], derived_key);
-
-        return { 'Ok': derived_key };
-    }
+    // async get_subkey_and_fetch_and_derive_if_needed(map_owner: Principal, map_name: string, map_key: string): Promise<{ 'Ok': CryptoKey } | { 'Err': string }> {
```
fspreiss, line 169: Also this can be removed, I guess.

fspreiss (approved):
Nice, thanks, @altkdf!

sdk/ic_vetkd_sdk_encrypted_maps_example/src/index.test.ts
```diff
@@ -158,19 +126,11 @@ test('get value should work', async () => {
 
   const set_value_result = await encrypted_maps.set_value(owner, "some map", "some key", value);
 
-  expect("Ok" in set_value_result).to.equal(true);
+  expect(set_value_result).toBeFalsy();
```
fspreiss, line 129: Hm...shouldn't this be true?

sdk/ic_vetkd_sdk_encrypted_maps/src/index.ts
altkdf: Ah, forgot about it, done!

sdk/ic_vetkd_sdk_encrypted_maps/src/index.ts
altkdf: Yes, done.

sdk/ic_vetkd_sdk_encrypted_maps_example/src/index.test.ts
altkdf: No:
* Some previous value -> `object`
* No previous value -> `undefined` (evaluates to `false`)
* Error during call -> `Error`

fspreiss: 👍 I see.

I asked because before this PR, we checked equality with true, so I assumed that true is returned if the value was successfully stored.

We may want to extend the documentation to say that the previous value, if any, is returned.

altkdf: I agree. I would prefer to do this after we change all the encrypted maps API to look more like JS to have a more stable API that we can write docs for. So basically how it works in my previous [reply](https://github.com/dfinity/vetkd-devkit/pull/52#discussion_r2016707511). Then, we definitely should add the docs.

PR #55, opened 2025-03-28
feat: CRP-2748 make encrypted maps and key manager API more JS-like 

altkdf:


altkdf:
This also contains changes from https://github.com/dfinity/vetkd-devkit/pull/49 to be merged after that. Github stacked PRs work in a weird way, so I will merge `main` once that PR is merged and the redundant changes will be gone.

altkdf:
Currently, we are returning `string` for things like map name or map key. But maybe it makes sense to return a `Uint8Array` and provide a function `bytesToString()`.

We generally could just use strings and not bytes underneath, but we use stable structure and map/key ids from tuples. There, we are not allowed to use variable-size types in the tuples. That means we need to set the buffer size in advance for such things and this extra storage overhead is not nice. That's why I think it makes sense to keep using bytes.

I'm not yet 100% sure about where which types should be (converted).

altkdf:
Something still doesn't work in the examples - I'll look at it tomorrow.

@fspreiss @randombit  any opinion on the changed API?

fspreiss (approved):
LGTM, as discussed via Zoom

PR #61, opened 2025-04-02
chore: CRP-2747 merge crates and packages

altkdf:
TODO description


altkdf:
This PR includes the changes from https://github.com/dfinity/vetkd-devkit/pull/60 and, therefore, should be merge afterwards.

.github/workflows/sdk_utils.yml
```diff
@@ -39,8 +48,18 @@ jobs:
...
           set -eExuo pipefail
-          cd frontend/ic_vetkd_sdk_utils
           npm install
+          pushd backend/canisters/ic_vetkeys_encrypted_maps_canister
+          dfx start --background
+          make mock
+          eval $(make export-cmd)
+          popd
+          pushd backend/canisters/ic_vetkeys_manager_canister
+          dfx start --background
```
fspreiss: Is there an advantage/disadvantage to doing `dfx start` twice vs. once for the whole project?

.github/workflows/backend.yml
fspreiss, line 1: There are a few `cdk` names left in this file.

.github/workflows/frontend.yml
fspreiss, line 1: There are a few `sdk` names left in this file.

frontend/ic_vetkeys/src/tools/index.ts
fspreiss, line 1: Having both a `tools` and a `utils` subfolder seems somewhat confusing. We likely want to improve this, but this can be done in a follow up PR once we know/agree on how to do that.

fspreiss (approved):
Thanks, @altkdf, for this extensive restructuring.

Looking at this, I'm still wondering if we should have the "core" utilities (TransportSecretKey, VetKey, EncryptedVetKey, etc.) in a separate package (e.g., `ic-vetkeys-core` or so). Anyway, this PR is clearly a big step in the right direction, so even if we decide to do further renamings, it seems they can relatively easily be done on top of this.

.github/workflows/backend.yml
altkdf: fixed in c283cfb99e9ce8020577b9bf3dbf19177f85fe70

.github/workflows/frontend.yml
altkdf: fixed in c283cfb99e9ce8020577b9bf3dbf19177f85fe70

frontend/ic_vetkeys/src/tools/index.ts
altkdf: Created [CRP-2770](https://dfinity.atlassian.net/browse/CRP-2770)

PR #65, opened 2025-04-08
feat: CRP-2757 implement Motoko backend

altkdf:


altkdf:
The map type operations currently incur much more overhead than they should and the reason is that the map types in motoko don't yet support range functions that we need for efficiency. As it looks, this limitation should soon be removed though https://github.com/dfinity/new-motoko-base/issues/271.

altkdf:
In next PRs: motoko-based canisters, better documentation, pocket IC tests checking that the VetKD calls work as intended.

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+        OrderedMap.Make<Caller>(Principal.compare);
+    };
+
+    func sharedKeysMapOps() : OrderedMap.Operations<KeyId> {
+        OrderedMap.Make<KeyId>(compareKeyIds);
+    };
+
+    // KeyManager class
+    public class KeyManager<T>(domainSeparator: Text, accessRightsOperations : Types.AccessControlOperations<T>) {
+        public var accessControl : OrderedMap.Map<Caller, [(KeyId, T)]> = accessControlMapOps().empty();
```
fspreiss, line 52: naming: I might have mentioned this earlier, but I find the name `permissions` would be a very good name for this map. In my view, _access control_ is the process of granting or denying a subject to perform an action (e.g., read, write) on a resource (e.g., an encrypted map entry) based on a set of permissions. And this map here captures exactly the permissions. I'm also OK with leaving it as is. And we could also do this in a separate PR (where we also rename the Rust-equivalent fields).

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+        public var sharedKeys : OrderedMap.Map<KeyId, [Caller]> = sharedKeysMapOps().empty();
+        public var managementCanisterPrincipalText = "aaaaa-aa";
+
+        // Get accessible shared key IDs for a caller
+        public func getAccessibleSharedKeyIds(caller : Caller) : [KeyId] {
```
fspreiss, line 57: I was wondering for a while why this method has "Shared" in its name: I though that `accessControl` includes the permissions for both the maps that are shared and the ones that are owned, but only realized later in an offline discussion with you that `accessControl` is really only capturing the permissions regarding the _sharings_, because owners can implicitly access their (owned) maps, so it is not necessary to capture this type of permission anywhere.

What would help is a brief documentation on the `accessControl` field (and on `sharedKeys`) that explains this.

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+                case (null) { [] };
+                case (?entries) {
+                    Array.map<(KeyId, T), KeyId>(entries, func((keyId, _)) = keyId);
+                };
+            };
+        };
+
+        // Get shared user access for a key
+        public func getSharedUserAccessForKey(caller : Caller, keyId : KeyId) : Result.Result<[(Caller, T)], Text> {
+            switch (ensureUserCanGetUserRights(caller, keyId)) {
```
fspreiss, line 68: nit/optional: how about doing `ensureUserCanGetUserRights` in an `if` and returning early if it doesn't hold? This would reduce the number of nestings in the function: currently there are 9 levels, which is a lot. If we do this, this would also apply to several other places.

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+                            #ok(Buffer.toArray(results));
+                        };
+                    };
+                };
+            };
+        };
+
+        // Get vetkey verification key
+        public func getVetkeyVerificationKey() : async VetKeyVerificationKey {
+            let derivationPath = [Text.encodeUtf8(domainSeparator)];
```
fspreiss, line 99: nit: how about saving the UTF8 encoding in a class variable, so that it only needs to be done once?

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+                    };
+
+                    let (reply) = await (actor (managementCanisterPrincipalText) : VetkdSystemApi).vetkd_derive_encrypted_key(request);
+                    #ok(reply.encrypted_key);
+                };
+            };
+        };
+
+        // Get user rights
+        public func getUserRights(caller : Caller, keyId : KeyId, user : Caller) : Result.Result<?T, Text> {
```
fspreiss, line 137: `user : Caller` doesn't seem correct, because the user isn't necessarily the caller: so, should it be `user : Principal`?

The same question applies to several other functions that have `user : Caller`.

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+            switch (ensureUserCanGetUserRights(caller, keyId)) {
+                case (#err(msg)) { #err(msg) };
+                case (#ok(_)) {
+                    #ok(
+                        do ? {
+                            if (Principal.equal(user, keyId.0)) {
+                                accessRightsOperations.ownerRights();
+                            } else {
+                                let entries = accessControlMapOps().get(accessControl, user)!;
```
fspreiss, line 146: Do I understand correctly that the `!` means that this directly returns `#ok(null)` if user doesn't have any rights?

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+                                let (k, rights) = Array.find<(KeyId, T)>(
+                                    entries,
+                                    func((k, rights)) = Principal.equal(k.0, keyId.0) and Blob.equal(k.1, keyId.1),
```
fspreiss, line 149: It might make sense to define a function `keyIdEquals` (or similarly named) and use that everywhere where the pattern `Principal.equal(k.0, keyId.0) and Blob.equal(k.1, keyId.1)` appears, which is at least 7 times in this file.

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+                        case (null) { [] };
+                        case (?users) { users };
+                    };
+
+                    let newUsers = switch (Array.indexOf<Caller>(user, currentUsers, Principal.equal)) {
+                        case (?index) {
+                            let mutCurrentUsers = Array.thaw<Caller>(currentUsers);
+                            mutCurrentUsers[index] := user;
+                            Array.freeze(mutCurrentUsers);
+                        };
```
fspreiss, line 179: Isn't this case a NO-OP: when indexOf user in currentUsers is X, then we set currentUsers[X] to user!?

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+                            let mutCurrentEntries = Array.thaw<(KeyId, T)>(currentEntries);
+                            oldRights := ?mutCurrentEntries[index].1;
+                            mutCurrentEntries[index] := (keyId, accessRights);
+                            Array.freeze(mutCurrentEntries);
+                        };
+                        case (null) {
+                            Array.append<(KeyId, T)>(currentEntries, [(keyId, accessRights)]);
+                        };
+                    };
+                    if (Array.size<(KeyId, T)>(newEntries) > 0) {
```
fspreiss, line 211: Isn't the size of `newEntries` always >=1? Above, either an existing entry is changed, or a new one is added, so `newEntries` should never be empty, no? Asked differently, do we really ever need to call line 214 (doing `.delete`)?

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+            if (Principal.equal(user, keyId.0)) {
+                return #ok(accessRightsOperations.ownerRights());
+            };
+
+            switch (accessControlMapOps().get(accessControl, user)) {
+                case (null) { #err("unauthorized") };
+                case (?entries) {
+                    for ((k, rights) in entries.vals()) {
+                        if (Principal.equal(k.0, keyId.0) and Blob.equal(k.1, keyId.1)) {
+                            return #ok(rights);
```
fspreiss, line 270: Here an implicit assumption is made that `canRead` always returns true. Is it really safe to assume that this is always the case? I'm thinking of a case where the accessRightsOperations are implemented so that someone might not always be allowed to read but only under certain conditions (e.g., time of day).

I'd say it would be safer to actually _call_ `canRead` here, rather than assuming that `canRead` will always returns true.
```
if (accessRightsOperations.canRead(rights)) {
    return #ok(rights);
 } else {
     return #err("unauthorized");
};
```

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
```diff
@@ -0,0 +1,311 @@
...
+import Types "../Types";
+import Text "mo:base/Text";
+import KeyManager "../key_manager/KeyManager";
+
+module {
+    type VetKeyVerificationKey = Blob;
+    type VetKey = Blob;
+    type Caller = Principal;
+    type KeyName = Blob;
+    type KeyId = (Caller, KeyName);
```
fspreiss, line 19: How about `(Owner, KeyName)` with type `Owner = Principal`, or alternatively `Creator`?

Also, maybe directly define MapId (i.e., `type MapId = (Owner, KeyName);`) and delete `type KeyId`, because `KeyId` shouldn't really be used in this file here (only in KeyManager.mo).

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
```diff
@@ -0,0 +1,311 @@
...
+    type MapName = KeyName;
+    type MapId = KeyId;
+    type MapKey = Blob;
+    type TransportKey = Blob;
+    type EncryptedMapValue = Blob;
+
+    type EncryptedMapData<T> = {
+        map_owner : Principal;
+        map_name : Blob;
```
fspreiss, line 28: nit: Blob -> MapName ?

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
```diff
@@ -0,0 +1,311 @@
...
+        keyvals : [(Blob, EncryptedMapValue)];
```
fspreiss, line 29: nit: Blob -> MapKey ?

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
```diff
@@ -0,0 +1,311 @@
...
+        return OrderedMap.Make<MapId>(compareMapIds);
+    };
+
+    public class EncryptedMaps<T>(domainSeparator : Text, accessRightsOperations : Types.AccessControlOperations<T>) {
+        public var keyManager = KeyManager.KeyManager<T>(domainSeparator, accessRightsOperations);
+        public var mapKeyVals : OrderedMap.Map<(MapId, MapKey), EncryptedMapValue> = mapKeyValsMapOps().empty();
+        public var mapKeys : OrderedMap.Map<MapId, [MapKey]> = mapKeysMapOps().empty();
+
+        // Get accessible shared map names for a caller
+        public func getAccessibleSharedMapNames(caller : Caller) : [MapId] {
```
fspreiss, line 68: This returns `[MapId]`, but the method is called `getAccessibleSharedMapNames`, i.e., it has `MapNames` in its name.

For comparison, in the key manager, the respective method is called `KeyManager::getAccessibleSharedKeyIds` (i.e., has `KeyIds` in its name) and returns `[KeyId]`.

So, should the naming be unified?

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
```diff
@@ -0,0 +1,311 @@
...
+                    };
+                };
+            };
+        };
+
+        // Get encrypted values for a map
+        public func getEncryptedValuesForMap(caller : Caller, mapId : MapId) : Result.Result<[(MapKey, EncryptedMapValue)], Text> {
+            switch (keyManager.getUserRights(caller, mapId, caller)) {
+                case (#err(msg)) { #err(msg) };
+                case (#ok(_)) {
```
fspreiss, line 108: Here it is implicitly assumed that if there are _any_ rights then this means there are "read" rights. But as I mentioned in my other comment, I'm not sure this is a valid assumption. I'd say here we should instead use an `ensureUserCan...` function.

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
```diff
@@ -0,0 +1,311 @@
...
+                    #ok(Buffer.toArray(values));
+                };
+            };
+        };
+
+        // Get encrypted value
+        public func getEncryptedValue(caller : Caller, mapId : MapId, key : MapKey) : Result.Result<?EncryptedMapValue, Text> {
+            switch (keyManager.getUserRights(caller, mapId, caller)) {
+                case (#err(msg)) { #err(msg) };
+                case (#ok(_)) {
```
fspreiss, line 131: Same as above: we should use an `ensureUserCan...` function.

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
```diff
@@ -0,0 +1,311 @@
...
+        public func getEncryptedVetkey(
+            caller : Caller,
+            mapId : MapId,
+            transportKey : TransportKey,
+        ) : async Result.Result<VetKey, Text> {
+            await keyManager.getEncryptedVetkey(caller, mapId, transportKey);
+        };
+
+        // Get user rights
+        public func getUserRights(caller : Caller, mapId : MapId, user : Caller) : Result.Result<?T, Text> {
```
fspreiss, line 282: `user : Caller` doesn't seem correct, because the user isn't necessarily the caller: so, should it be `user : Principal`?

The same question applies to `func removeUser`.

fspreiss (commented):
Thanks, @altkdf! Looks generally good! Please find a few comments below.

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
altkdf: Yes, this code is from the time before we added more flexible access control. This needs changes in Rust as well. I changed `ensureUserCanRead`, but let's revisit this in another PR. Created [CRP-2784](https://dfinity.atlassian.net/browse/CRP-2784).

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
altkdf: I slightly simplified it, but Motoko is not very flexible with `Result`. I guess the most problematic part IMO is that Motoko is not good with inferring types, which adds a lot of boilerplate code like `switch`es. So we could look at how to simplify this code (maybe later), but I'm not sure how far we will get and how much time this will cost us. Maybe we could ask the motoko team to look at this code, but I'm not sure how much time they currently have considering their priorities.

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+import Types "../Types";
+import Text "mo:base/Text";
+
+module {
+    type VetKeyVerificationKey = Blob;
+    type VetKey = Blob;
+    type Creator = Principal;
+    type Caller = Principal;
+    type KeyName = Blob;
+    type KeyId = (Caller, KeyName);
```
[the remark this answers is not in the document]
altkdf: done

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
```diff
@@ -0,0 +1,327 @@
...
+    };
+
+    func sharedKeysMapOps() : OrderedMap.Operations<KeyId> {
+        OrderedMap.Make<KeyId>(compareKeyIds);
+    };
+
+    // KeyManager class
+    public class KeyManager<T>(domainSeparator: Text, accessRightsOperations : Types.AccessControlOperations<T>) {
+        public var accessControl : OrderedMap.Map<Caller, [(KeyId, T)]> = accessControlMapOps().empty();
+        public var sharedKeys : OrderedMap.Map<KeyId, [Caller]> = sharedKeysMapOps().empty();
```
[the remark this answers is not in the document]
altkdf: done

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
altkdf: Since this would require changes in Rust as well, let's address this separately. Created [CRP-2782](https://dfinity.atlassian.net/browse/CRP-2782).

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
altkdf: Added a comment to [CRP-2676](https://dfinity.atlassian.net/browse/CRP-2676). More documentation will be added in a separate PR.

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
altkdf: done

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
altkdf: yes, fixed

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
altkdf: yes

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
altkdf: changed to `compareKeyIds(k, keyId) == #equal`

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
altkdf: yes, fixed

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
altkdf: yes, fixed

backend/mo/ic_vetkeys/src/key_manager/KeyManager.mo
altkdf: Yes, makes sense, done

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
altkdf: changed to `public type MapId = KeyManager.KeyId`.

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
altkdf: done

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
altkdf: This is the case in rust as well. This should be changed at some point. Created [CRP-2783](https://dfinity.atlassian.net/browse/CRP-2783) to track this.

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
altkdf: See https://github.com/dfinity/vetkd-devkit/pull/65#discussion_r2057827813

backend/mo/ic_vetkeys/src/encrypted_maps/EncryptedMaps.mo
altkdf: done

backend/mo/ic_vetkeys/test/EncryptedMaps.test.mo
```diff
@@ -0,0 +1,258 @@
...
+
+        // Give read access and verify still can't write
+        switch (encryptedMaps.setUserRights(p1, (p1, mapName), p2, #Read)) {
+            case (#ok(_)) {};
+            case (#err(e)) { Debug.trap("Failed to set read access: " # e) };
+        };
+
+        assert encryptedMaps.insertEncryptedValue(p2, (p1, mapName), mapKey, mapValue) == #err("unauthorized");
+    },
+);
```
fspreiss, line 161: nit: not sure, but maybe this test case can be integrated into the existing "unauthorized cannot invoke operations" test, which is also about ensuring that unauthorized access fails.

backend/mo/ic_vetkeys/test/EncryptedMaps.test.mo
```diff
@@ -0,0 +1,258 @@
...
+        switch (encryptedMaps.setUserRights(p1, (p1, mapName), p2, #Read)) {
+            case (#ok(_)) {};
+            case (#err(e)) { Debug.trap("Failed to set read access: " # e) };
+        };
+
+        // Add p2 access rights
+        switch (encryptedMaps.setUserRights(p1, (p1, mapName), p2, #ReadWriteManage)) {
+            case (#ok(_)) {};
+            case (#err(e)) { Debug.trap("Failed to set read access: " # e) };
+        };
```
fspreiss, line 205: Is it intentional that p2 gets access rights twice: first #Read and then #ReadWriteManage?

backend/mo/ic_vetkeys/test/EncryptedMaps.test.mo
```diff
@@ -0,0 +1,258 @@
...
+
+        // Modify value
+        switch (encryptedMaps.insertEncryptedValue(p1, (p1, mapName), mapKey, newValue)) {
+            case (#ok(?_)) {};
+            case (unexpected) {
+                Debug.trap("Failed to modify value: " # debug_show (unexpected));
+            };
+        };
+    },
```
fspreiss, line 236: Would it make sense to check that the entry at the end has the correct (=new) value?

backend/mo/ic_vetkeys/test/KeyManager.test.mo
```diff
@@ -0,0 +1,239 @@
...
+                Debug.trap("should set user rights: " # e);
+            };
+        };
+
+        switch (keyManager.getUserRights(p2, (p1, keyName), p2)) {
+            case (#err e) {
+                assert e == "unauthorized";
+            };
+            case (unexpected) {
+                Debug.trap("user should not have access rights withh only read rights " # debug_show (unexpected));
```
fspreiss, line 86: withh -> with

backend/mo/ic_vetkeys/test/KeyManager.test.mo
```diff
@@ -0,0 +1,239 @@
...
+
+            switch (keyManager.removeUserRights(remover, (p1, keyName), p2)) {
+                case (#ok arg) {
+                    assert arg == ?#ReadWriteManage;
+                };
+                case (unexpected) {
+                    Debug.trap("unexpected result in removing user rights: " # debug_show (unexpected));
+                };
+            };
+        };
```
fspreiss, line 143: Maybe also check again that now the user really doesn't have any rights anymore?

backend/mo/ic_vetkeys/test/KeyManager.test.mo
```diff
@@ -0,0 +1,239 @@
+import VetKey "../src";
```
fspreiss, line 1: Should we also add tests for `getVetkeyVerificationKey` and `getEncryptedVetkey`?

fspreiss (approved):
Thanks for addressing the comments and for creating the tickets for the remaining ones, @altkdf!

backend/mo/ic_vetkeys/test/EncryptedMaps.test.mo
altkdf: done

backend/mo/ic_vetkeys/test/EncryptedMaps.test.mo
altkdf: No, we should iterate through access rights instead. Fixing it also revealed a bug that I now also fixed. See ded4210.

backend/mo/ic_vetkeys/test/EncryptedMaps.test.mo
altkdf: added

backend/mo/ic_vetkeys/test/KeyManager.test.mo
altkdf: done

backend/mo/ic_vetkeys/test/KeyManager.test.mo
altkdf: done

backend/mo/ic_vetkeys/test/KeyManager.test.mo
altkdf: I'm not sure what would be a meaningful for of doing this without creating a canister that calls it. This would be essentially a mock checking that the arguments are passed in a correct way, so essentially building the same code a second time and adding some complexity. So my current preference is to test this from a canister. Since we have Rust tests already, we could just reuse them because they are using rust tests + pocket IC + wasm files, which works for motoko as well.

PR #66, opened 2025-04-10
chore: use sized principal bytes in derivation id

altkdf:


backend/ic_vetkeys/src/key_manager/mod.rs
```diff
@@ -290,6 +282,14 @@ fn vetkd_system_api_canister_id() -> CanisterId {
     CanisterId::from_str(VETKD_SYSTEM_API_CANISTER_ID).expect("failed to create canister ID")
 }
 
+pub fn key_id_to_derivation_id(principal: Principal, key_name: &[u8]) -> Vec<u8> {
+    let mut derivation_id = Vec::with_capacity(principal.as_slice().len() + 1 + key_name.len());
+    derivation_id.push(principal.as_slice().len() as u8);
+    derivation_id.extend(principal.as_slice());
+    derivation_id.extend(key_name);
+    derivation_id
```
fspreiss, line 290: I think we could even consider using a domain separator here...

fspreiss (approved):

backend/ic_vetkeys/src/key_manager/mod.rs
altkdf: Discussed offline: the `derivation_path` (soon to be changed in this repo to `context` as is done upstream) is instantiated in the constructor and already serves as domain separator, so it doesn't seem that an additional one here would bring an improvement.

PR #69, opened 2025-04-11
feat: CRP-2775 add a simple IBE example

altkdf:
Implements simple encrypted messaging between `Principal`s.


altkdf:
A readme still needs to be added. Maybe actually in another PR.

examples/basic_ibe/src/declarations/basic_ibe/basic_ibe.did
```diff
@@ -0,0 +1,19 @@
+type Inbox = record { messages : vec Message };
+type Message = record {
```
randombit, line 2: IIUC the files in `declarations` are all generated. Should they still be checked into git? (I don't know what the convention is for this regarding canisters)

examples/basic_ibe/backend/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+        } else {
+            inbox.messages.push(message);
+            inboxes.insert(receiver, inbox);
+            Ok(())
+        }
+    })
+}
+
+#[update]
+async fn get_root_ibe_public_key() -> VetKeyPublicKey {
```
randombit, line 62: What is this used for?

examples/basic_ibe/backend/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+#[query]
+fn get_my_messages() -> Inbox {
+    let caller = ic_cdk::caller();
+    INBOXES.with_borrow_mut(|inboxes| inboxes.get(&caller).unwrap_or_default())
+}
+
+#[update]
+fn remove_my_messages() -> Inbox {
```
randombit, line 108: Maybe ok for the purposes of an example but this is not atomic and can lose messages if interleaved. Eg

Time 1) Inbox contains messages `a` and `b`
Time 2) User calls get_my_messages() returning `a` and `b`
Time 3) Someone sends message `c`
Time 4) User decrypts `a` and `b`, then calls `remove_my_messages`

This doesn't have to be addressed but at least a comment calling out the issue would be nice so someone knows if they copy this example they have to deal with such interleavings.

As to addressing, possible solutions:

1) `get_my_messages` returns and deletes the messages (but this fails if the response doesn't make it all the way to the user, eg because the BN fails after the request is received but before the response is sent)
2) `remove_my_messages` takes some input indicating which messages are safe to delete
 2a) A list of unique message identifiers to remove (eg SHA-256 of the ciphertext)
 2b) A timestamp; all messages sent before this time are removed
 2c) A count; the first `n` messages are removed. [This assumes messages are appended in order, which seems to currently be the case]

examples/basic_ibe/backend/src/types.rs
```diff
@@ -0,0 +1,104 @@
...
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+use std::borrow::Cow;
+
+pub const MAX_MESSAGES_PER_INBOX: usize = 10;
+
+#[derive(CandidType, Serialize, Deserialize, Clone, Debug)]
+pub struct Message {
+    pub sender: Principal,
+    pub receiver: Principal,
```
randombit, line 13: IIUC the receiver could be sent once as part of the `Inbox`, since all messages in an inbox are associated with that single receiver.

examples/basic_ibe/backend/src/types.rs
```diff
@@ -0,0 +1,104 @@
...
+#[derive(CandidType, Deserialize)]
+pub(crate) struct VetKDPublicKeyRequest {
+    pub canister_id: Option<CanisterId>,
+    #[serde_as(as = "Vec<serde_with::Bytes>")]
+    pub derivation_path: Vec<Vec<u8>>,
+    pub key_id: VetKDKeyId,
+}
+
+#[derive(CandidType, Deserialize)]
+pub(crate) struct VetKDPublicKeyReply {
```
randombit, line 83: The VetKD definitions aren't in the CDK are they :(

Should there be / is there a ticket for this? Ideally this is fixed before launch...

examples/basic_ibe/backend/src/types.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub public_key: Vec<u8>,
+}
+
+#[serde_as]
+#[derive(CandidType, Deserialize, Clone, Debug)]
+pub(crate) struct VetKDEncryptedKeyRequest {
+    #[serde_as(as = "Vec<serde_with::Bytes>")]
+    pub public_key_derivation_path: Vec<Vec<u8>>,
+    #[serde(with = "serde_bytes")]
+    pub derivation_id: Vec<u8>,
```
randombit, line 94: Should in our new naming this be called `input`?

examples/basic_ibe/src/declarations/basic_ibe/basic_ibe.did
altkdf: It's not necessary to include them, but we did it in this repo so far. The advantage I see is that you don't need to run any `npm` commands to see all files, but we may want to change it because we either way need to run `npm install` for static analysis and maybe something else as well. Created [CRP-2780](https://dfinity.atlassian.net/browse/CRP-2780).

examples/basic_ibe/backend/src/lib.rs
altkdf: This function also returns the removed `Inbox`, so if a message `c` has arrived and the caller didn't see it yet, this function will return `c` but also `a' and `b` and so no messages would be lost. This may not be very efficient for large applications, but that's also not the purpose of this example.

examples/basic_ibe/backend/src/types.rs
altkdf: Yes, once the chainkey testing canister changes to the API are released, we can rename it.

altkdf:
I now restyled the messages, especially with respect to their deletion, and renamed the deletion button to be less confusing.

Screenshot From 2025-04-16 08-41-10
Screenshot From 2025-04-16 08-41-41

examples/basic_ibe/frontend/src/typescript.svg
fspreiss, line 1: Do we need the typescript logo checked in? I don't see it displayed in the app anywhere.

fspreiss (approved):
Very nice example, thanks, @altkdf! To simplify the app even further you could consider doing either of the following:
1. Adapting the UI so that upon pressing the Delete button, only the "accidentally deleted" messages are show, i.e., the ones that were already displayed are not shown again.
2. Have a delete button per message instead of the Delete-All button. But thinking about this again (after discussing it in the team meeting), when deleting messages by index, we have to be careful to delete the correct message in case new messages arrived on the server that are not yet know in the frontend.

examples/basic_ibe/frontend/src/typescript.svg
altkdf: No, removed.

examples/basic_ibe/frontend/public/vite.svg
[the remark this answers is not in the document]
altkdf: yes, that's the website logo

examples/basic_ibe/backend/backend.did
```diff
@@ -0,0 +1,18 @@
...
+  timestamp : nat64;
+  encrypted_message : blob;
+};
+type Result = variant { Ok; Err : text };
+type SendMessageRequest = record {
+  encrypted_message : blob;
+  receiver : principal;
+};
+service : {
+  get_my_encrypted_ibe_key : (blob) -> (blob);
```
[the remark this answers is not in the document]
altkdf: That would require editing candid manually, while we're generating it automatically. So after generating it again we should always not forget to change it manually. I would prefer to keep it as is if that's okay with you.

examples/basic_ibe/backend/src/types.rs
```diff
@@ -0,0 +1,103 @@
+use candid::{CandidType, Principal};
+use ic_cdk::api::management_canister::main::CanisterId;
+use ic_stable_structures::{storable::Bound, Storable};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+use std::borrow::Cow;
+
+pub const MAX_MESSAGES_PER_INBOX: usize = 10;
```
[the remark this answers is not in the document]
altkdf: Increased to 1000.

altkdf:
I replaced the current delete all approach with a delete button for each message.

PR #74, opened 2025-04-25
feat: CRP-2762 add a timelock encryption example (secret bid auction)

altkdf:
Adds a basic example for identity based timelock encryption. The use case is a dummy secret bid auction, where any authenticated principal can create a lot with a name, description, and a deadline, and any authenticated principal but the creator of the lot can place a bid with a dummy value, i.e., there is neither real assets involved nor is there any real limitation except the value being a `u128`.

The bids are encrypted towards the lot id (generated by the canister) and are decrypted by a timer job that is running every 5 seconds and taking the oldest expired lot for decryption.

Also, this PR bumps the commit of `ic-vetkd-utils` that was required to fix IBE decryption inside the canister (because it was using an outdated ciphertext format).


altkdf:
It seems that the IBE ciphertext serialization is inconsistent between js and rust. This PR likely needs to wait until it's fixed. See [CRP-2785](https://dfinity.atlassian.net/browse/CRP-2785).

altkdf:
The readme and eventually more documentation will be added in a separate PR.

examples/basic_timelock_ibe/backend/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+                .map(|(_, bid)| match bid {
+                    Bid::Encrypted(encrypted_bid) => encrypted_bid,
+                    Bid::Decrypted(_) => panic!("bug: decrypted bid in a closed lot"),
+                })
+                .collect()
+        });
+
+        let decrypted_bids = decrypt_bids(lot_id, encrypted_bids, root_ibe_public_key).await;
+
+        let status = match decrypted_bids.iter().max_by(|x, y| x.amount.cmp(&y.amount)) {
```
randombit, line 227: If multiple users bid the same (highest) amount then the last user to bid wins, which seems weird. [*]

I think just using `iter().rev()` resolves this since then `max_by` returns the last (ie earliest) maximum bid.

[*] In some models it could even be a security issue because the encrypted bid is just the integer with no context binding. So if I am in privileged position (ie running a node in the subnet) and can see the encrypted bids as they come in, I can win the auction always with minimum overhead by copying each users bid and sending each bid again as myself. Since I am the last bidder to make the highest bid, I win.

randombit: Should a bid of 0 be accepted?

examples/basic_timelock_ibe/backend/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+            ..
+        }) if creator != bidder && now < end_time => Ok(()),
+        Some(LotInformation { creator, .. }) if creator == bidder => {
+            Err("lot creator cannot bid".to_string())
+        }
+        Some(_) => Err("lot is closed".to_string()),
+        None => Err("lot not found".to_string()),
+    })?;
+
+    if encrypted_amount.len() > 1000 {
```
randombit, line 169: Since the plaintext length is (IIUC) fixed to 16 bytes, this check could be more precise, even exactly checking for the expected size. [Maybe a bit fragile if we change something in the future about the IBE encryption, though]

examples/basic_timelock_ibe/backend/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+            creator,
+            end_time,
+            ..
+        }) if creator != bidder && now < end_time => Ok(()),
+        Some(LotInformation { creator, .. }) if creator == bidder => {
+            Err("lot creator cannot bid".to_string())
+        }
+        Some(_) => Err("lot is closed".to_string()),
+        None => Err("lot not found".to_string()),
+    })?;
```
randombit, line 167: IIUC the same bidder can make multiple bids. Which is maybe OK/expected; their highest bid is what counts in the end. I'm just thinking that `close_one_lot_if_any_is_open` may run into problems if there are millions of bids on one lot, and decrypting the lot takes more than 5 seconds. [I don't know how the runtime reacts to this situation, where a timer goes off and the function is still running when the next timer round occurs]

examples/basic_timelock_ibe/backend/src/lib.rs
altkdf: The goal was to limit that to some extend, but we could change format of the plaintext or ciphertext, so that was a very generous limit. If you have some better/more precise constant, we could change the current one.

examples/basic_timelock_ibe/backend/src/lib.rs
altkdf: A bid of 0 doesn't make much sense to me/don't know what the use case would be :thinking: 

Regarding the equal bids, we could just call `raw_rand` and decide based on that?

altkdf: > Regarding the equal bids, we could just call raw_rand and decide based on that?

But maybe letting the first bid win is a better strategy because it doesn't have an additional latency.

altkdf: Changed in c925ccc2ed47fdc6dca8889ea7ff42daf6f2c695.

examples/basic_timelock_ibe/backend/src/lib.rs
randombit: > We overwrite user's bid if passed multiple ones and only the last one counts. So you would need many principals for that.

OK I had missed this part. I thought if a single principal spammed orders we would not dedup.

examples/basic_timelock_ibe/backend/src/lib.rs
randombit: IBE overhead currently is

* 8 byte header/version tag
* 48 bytes G1 point
* 32 bytes seed
* The masked ciphertext (without any padding or auth tag, etc)

So a 16 byte plaintext will always (at the moment) encrypt to a 104 byte ciphertext.

Fine to also just be inexact especially in an example.

randombit (approved):

examples/basic_timelock_ibe/backend/src/lib.rs
altkdf: Thinking about this again, I would be for adding a more precise check if we would provide a devkit function like `ibe_ciphertext_size(plaintext_size: usize) -> usize`. This seems to be a better option than hard-coding a constant here because we would be basically giving the user responsibility to get things right. Created [CRP-2787](https://dfinity.atlassian.net/browse/CRP-2787) to add the check. And will add a link to this comment.

examples/basic_timelock_ibe/backend/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+        });
+
+        let decrypted_bids = decrypt_bids(lot_id, encrypted_bids, root_ibe_public_key).await;
+
+        let status = match decrypted_bids
+            .iter()
+            .rev() // reverse the bids to get the *oldest* maximum bid
+            .max_by(|x, y| x.amount.cmp(&y.amount))
+        {
+            Some(winner_bid) => LotStatus::ClosedWithWinner(winner_bid.bidder),
```
fspreiss, line 232: As discussed offline, this doesn't actually return the _oldest_ maximum bid.

Thanks for the fix in https://github.com/dfinity/vetkd-devkit/pull/89.

PR #75, opened 2025-04-29
Address an informational finding from the internal security review

randombit:
The Rust library uses the name `header` for this parameter, which more accurately captures what the value is, and it's best if the Rust and TS libraries are consistent whenever possible.


altkdf (approved):
LGTM! Maybe also replace `hdr` with `header` to keep only one style of writing.

randombit:
I went in the other direction (`header` everywhere), `hdr` is maybe a bit more clipped than truly necessary here...

PR #81, opened 2025-05-02
chore: CRP-2742 add prettier and linting to `ic_vetkeys`

altkdf:


fspreiss (approved):
Thanks, @altkdf! Maybe add a PR description saying that all changes in *.ts files are purely formatting-related (at least that was my impression when looking at the files).

Note that there are also still a few merge conflicts.

PR #82, opened 2025-05-05
feat: CRP-2757 add key manager canister for motoko

altkdf:


backend/mo/canisters/ic_vetkeys_manager_canister/src/Main.mo
```diff
@@ -0,0 +1,81 @@
...
+        #Err : Err;
+    };
+
+    public query (msg) func get_accessible_shared_key_ids() : async [(Principal, Blob)] {
+        keyManager.getAccessibleSharedKeyIds(msg.caller);
+    };
+
+    public query (msg) func get_shared_user_access_for_key(
+        key_owner : Principal,
+        key_name : Blob,
```
fspreiss, line 22: The respective Rust canister takes `key_name` as `ByteBuf`, while here it is a `Blob`. Shouldn't this be the same?

backend/mo/canisters/ic_vetkeys_manager_canister/src/Main.mo
```diff
@@ -0,0 +1,81 @@
...
+import Blob "mo:base/Blob";
+import Result "mo:base/Result";
+
+actor {
+    var keyManager = KeyManager.KeyManager<Types.AccessRights>("key_manager", Types.accessRightsOperations());
+    public type ByteBuf = { inner : Blob };
+    public type Result<Ok, Err> = {
+        #Ok : Ok;
+        #Err : Err;
+    };
```
fspreiss, line 14: Do I understand correctly that the only difference to mo:base/Result.Result is that here the variant names are capitalized?

It unfortunate that such a conversion is even necessary...

backend/mo/canisters/ic_vetkeys_manager_canister/src/Main.mo
```diff
@@ -0,0 +1,81 @@
...
+
+    public query (msg) func get_shared_user_access_for_key(
+        key_owner : Principal,
+        key_name : Blob,
+    ) : async Result<[(Principal, Types.AccessRights)], Text> {
+        convertResult(keyManager.getSharedUserAccessForKey(msg.caller, (key_owner, key_name)));
+    };
+
+    public shared func get_vetkey_verification_key() : async ByteBuf {
```
fspreiss, line 27: Here, both the Rust canister and the Motoko canister return a ByteBuf, so the same type is used, while for other methods the Motoko canister sometimes specifies `Blob` instead of `ByteBuf`.

backend/mo/canisters/ic_vetkeys_manager_canister/src/Main.mo
```diff
@@ -0,0 +1,81 @@
...
+    public shared (msg) func get_encrypted_vetkey(
+        key_owner : Principal,
+        key_name : ByteBuf,
+        transport_key : ByteBuf,
+    ) : async Result<ByteBuf, Text> {
+        let vetkeyBytebuf = await keyManager.getEncryptedVetkey(msg.caller, (key_owner, key_name.inner), transport_key.inner);
+        switch (vetkeyBytebuf) {
+            case (#err(e)) { #Err(e) };
+            case (#ok(inner)) { #Ok({ inner }) };
+        };
```
fspreiss, line 41: Could we use `convertResult(vetkeyBytebuf)` here?

backend/mo/canisters/ic_vetkeys_manager_canister/src/Main.mo
```diff
@@ -0,0 +1,81 @@
+import { KeyManager } "../../../ic_vetkeys/src";
+import Types "../../../ic_vetkeys/src/Types";
+import Principal "mo:base/Principal";
+import Text "mo:base/Text";
+import Blob "mo:base/Blob";
+import Result "mo:base/Result";
+
+actor {
+    var keyManager = KeyManager.KeyManager<Types.AccessRights>("key_manager", Types.accessRightsOperations());
+    public type ByteBuf = { inner : Blob };
```
fspreiss, line 10: I forgot, did we document somewhere in the repo why this ByteBuf-blob-wrapper is sometimes used in the API, rather than just a blob?

backend/mo/canisters/ic_vetkeys_manager_canister/Makefile
```diff
@@ -0,0 +1,11 @@
...
+
+.PHONY: compile-wasm
+.SILENT: compile-wasm
+compile-wasm:
+	dfx build --check
+
+.PHONY: test
+.SILENT: test
+test: compile-wasm
+	CUSTOM_WASM_PATH=$(PWD)/.dfx/local/canisters/ic_vetkeys_manager_canister/ic_vetkeys_manager_canister.wasm cargo test -p ic-vetkeys-manager-canister
```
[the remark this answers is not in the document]
altkdf: added

backend/mo/canisters/ic_vetkeys_manager_canister/src/Main.mo
```diff
@@ -0,0 +1,81 @@
...
+    public type Result<Ok, Err> = {
+        #Ok : Ok;
+        #Err : Err;
+    };
+
+    public query (msg) func get_accessible_shared_key_ids() : async [(Principal, Blob)] {
```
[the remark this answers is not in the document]
altkdf: All APIs should return `ByteBuf`, this was an oversight. Fixed.

backend/mo/canisters/ic_vetkeys_manager_canister/src/Main.mo
altkdf: correct

backend/mo/canisters/ic_vetkeys_manager_canister/src/Main.mo
altkdf: No, we are also converting between `Blob` and `ByteBuf` in-place. I tried to first do that conversion, and then do the `Result` conversion, but that turned out to result in a longer code. So I kept the initial version.

backend/mo/canisters/ic_vetkeys_manager_canister/src/Main.mo
altkdf: Added a comment there and to the respective Rust `ByteBuf`.

backend/mo/canisters/ic_vetkeys_manager_canister/src/Main.mo
fspreiss: I think developers would find it useful to have some documentation for our own `Result` type explaining why this is used/needed in addition to the already existing `mo:base/Result`.

fspreiss (approved):
Thanks, @altkdf!

backend/mo/canisters/ic_vetkeys_manager_canister/src/Main.mo
altkdf: added a comment

PR #85, opened 2025-05-06
feat: CRP-2762 add a README for `basic_timelock_ibe` example

altkdf:


examples/basic_timelock_ibe/README.md
```diff
@@ -0,0 +1,79 @@
+# Basic Timelock Identity Based Encryption
+
+> [!IMPORTANT]  
+> These support libraries are under active development and are subject to change. Access to the repositories have been opened to allow for early feedback. Please check back regularly for updates.
+
+The **Basic Timelock IBE** example demonstrates how to use **[VetKeys](https://internetcomputer.org/docs/building-apps/network-features/encryption/vetkeys)** to implement secret-bid auction using Identity Based Encryption (IBE) on the **Internet Computer (IC)**. This application allows users authenticated using their **Internet Identity Principal** to create auction lots with a description and deadline and other users to place a secret bid for the lot. The bids in this example are just dummy integer values, contrary to real-world use cases where users would place bids holding some value.
```
fspreiss, line 6: > The bids in this example are just dummy integer values, contrary to real-world use cases where users would place bids holding some value.

Just to be sure: what you are trying to say is that the example dapp doesn't require real money, right?

examples/basic_timelock_ibe/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+This canister (IC smart contract) ensures that:
+1. Only the authorized user can create auction lots and place secret bids until the lot is closed.
```
fspreiss, line 9: > Only the authorized user

This sounds as if only a single user is authorized. Do you mean to say "Only authorized users can..."

examples/basic_timelock_ibe/README.md
```diff
@@ -0,0 +1,79 @@
...
+The code of smart contract determines the rules for when the decryption happens.
+In this example, the decryption happens only after the lot is closed and does not accept new bids.
+A canister functionality for decrypting secrets can be detected by inspecting the code and, therefore, it is crucial that canisters using VetKeys have their code public to allow to verify that the canister handles secrets in a secure and intended way.
+
+![UI Screenshot](ui_screenshot.png)
+
+## Features
+
+- **Secret Bid Placement**: Uses IBE capabilities of IC Vetkeys to encrypt messages that can only be decrypted by the intended recipient.
+- **Bid Id Based Encryption**: Each bid gets a unique bid id, and the secret bids are encrypted to the bid id as the public key identifier.
```
fspreiss, line 23: Maybe say "BidId-based Encryption" or "Bid-Id-based Encryption"?

examples/basic_timelock_ibe/README.md
```diff
@@ -0,0 +1,79 @@
...
+- **Time-Based Access Control**: Messages can only be decrypted after a specified time period has elapsed.
+
+## Efficiency
+
+- **Reducing the Number of VetKeys for Decryption**: In the current implementation, the bids are encrypted with the unique public key corresponding to the bid id. This is not strictly necessary and can be optimized if there is a need. For example, the bids could be decrypted based on a timestamp, e.g., every minute. All bids that were closed in that minute would be decrypted with one VetKey. The required change would be to encrypt bids with a predefined timestamp as the public key identifier instead of the bid id.
+- **Public Key Retrieval**: It is possible to use one public key to encrypt bids for multiple auction lots. The subkey derivation for a concrete lot can happen on the client side, i.e., in the frontend. It is already the case in this example. A further, minor optimization is possible by hardcoding the root public key of the IC, removing the need and the latency for obtaining the public key from the IC.
```
fspreiss, line 29: minor: Here it would be useful for readers to know where to get "the root public key of the IC" from (I guess we'll have that in the documentation at some point; or do you mean the canister master public key?), where it should be hardcoded (backend, frontend), and how to do the derivation (could we do this directly in the example?). Not sure if we really want to address all this now, especially because the optimization would be very minor, but at least these where the questions that came up for me when I read the sentence.

examples/basic_timelock_ibe/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+Note that currently in a local deployment, the [chainkey *testing* canister](https://github.com/dfinity/chainkey-testing-canister) is used to mock the VetKD protocol.
+
+## Example Components
+
+### Backend
+
+The backend consists of a canister that:
+* Lets users create auction lots with a description and duration.
+* Stores at most one encrypted bid from any authenticated user but the creator of the lot. Secret bids failing do decrypt are ignored. If a user provides multiple bids, only the last one is considered. The ciphertexts for secret bids of unexpectedly large size are rejected. Bids to expired lots are rejected.
```
fspreiss, line 59: > from any authenticated user but the creator of the lot

How about "from any authenticated user _except_ the creator of the lot"?

examples/basic_timelock_ibe/README.md
```diff
@@ -0,0 +1,79 @@
...
+* Allows users to retrieve the status of the lot, including the winner and the decrypted bids once the lot is closed.
+* A timer inside the canister periodically runs and takes one closed lot that it decrypts. if multiple users provide the same bids, the bid that was placed first becomes the winner.
```
fspreiss, line 61: > if multiple users

"If" should be capitalized

fspreiss, line 61: > if multiple users provide the same bids

Do you mean the same "highest" bid?

> the bid that was placed first becomes the winner

Or just: "the bid that was placed first wins."

fspreiss (commented):
Thanks, @altkdf! Nice README! Here are some minor comments.

examples/basic_timelock_ibe/README.md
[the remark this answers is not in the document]
altkdf: done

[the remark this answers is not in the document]
altkdf: done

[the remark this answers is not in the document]
altkdf: done

altkdf: yes

examples/basic_timelock_ibe/README.md
altkdf: done

examples/basic_timelock_ibe/README.md
altkdf: done

examples/basic_timelock_ibe/README.md
altkdf: The comment is about the subnet root key. Should we remove it if it's a minor optimization?

examples/basic_timelock_ibe/README.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+### Deploy the Canisters
+
+Run the local deployment script, which starts the local development environment (`dfx`) if necessary, builds both backend and frontend (asset) canisters, and installs them locally.
+```bash
+bash deploy_locally.sh
+```
+
+Note that currently in a local deployment, the [chainkey *testing* canister](https://github.com/dfinity/chainkey-testing-canister) is used to mock the VetKD protocol.
```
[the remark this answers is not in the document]
altkdf: added

examples/basic_timelock_ibe/README.md
altkdf: done

examples/basic_timelock_ibe/README.md
altkdf: done

altkdf: done

examples/basic_timelock_ibe/README.md
```diff
@@ -0,0 +1,79 @@
...
+> These support libraries are under active development and are subject to change. Access to the repositories have been opened to allow for early feedback. Please check back regularly for updates.
+
+The **Basic Timelock IBE** example demonstrates how to use **[VetKeys](https://internetcomputer.org/docs/building-apps/network-features/encryption/vetkeys)** to implement a secret-bid auction using Identity Based Encryption (IBE) on the **Internet Computer (IC)**. This application allows users authenticated with their **Internet Identity Principal** to create auction lots with a description and deadline, and other users to place a secret bid for the lot. The bids in this example are just dummy integer values, contrary to real-world use cases where users would place bids holding some value.
+
+This canister (IC smart contract) ensures that:
+1. Only authorized users can create auction lots and place secret bids until the lot is closed.
+2. The bids stay secret until the lot is closed.
+3. The winner is chosen fairly among all the placed bids, once the lot closes and the canister decrypts the secret bids. Note that once secret bids are decrpyted they inherently become public.
```
randombit, line 11: typo: decrypted

examples/basic_timelock_ibe/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+Note that currently in a local deployment, the [chainkey *testing* canister](https://github.com/dfinity/chainkey-testing-canister) is used to mock the VetKD protocol.
+
+## Example Components
+
+### Backend
+
+The backend consists of a canister that:
+* Lets users create auction lots with a description and duration.
+* Stores at most one encrypted bid from any authenticated user except the creator of the lot. Secret bids failing do decrypt are ignored. If a user provides multiple bids, only the last one is considered. The ciphertexts for secret bids of unexpectedly large size are rejected. Bids to expired lots are rejected.
```
randombit, line 59: failing *to* decrypt

examples/basic_timelock_ibe/README.md
altkdf: done

examples/basic_timelock_ibe/README.md
altkdf: done

examples/basic_timelock_ibe/README.md
fspreiss: Actually, no matter which public key is meant, only a single public key needs to be transmitted from the backend to the frontend (and optionally the frontend then needs to do some further derivation, or not, depending on whether the derivation was already done in the backend).

And the example dapp is already implemented in the optimal way, i.e., so that just a single key is transmitted, because `getRootIbePublicKey` caches the key (i.e., it doesn't fetch it again if it has it already).

So unless I'm missing something, I don't think this can actually be further optimized: even if a key is "hard-coded", this hard-coded key still needs to be transmitted to the client (e.g., when the frontend is loaded for the first time).

So, I'd just remove this bullet point.

altkdf: We could encode the subnet public key and derive everything we need without any actual interaction with the canister. I meant that. But I could still remove it if it's a minor improvement.

altkdf: Replaced with
> Currently, the backend canister, upon first usage of the IBE public key, is making a call to the management canister to fetch this root public key. This key is then cached in the canister. The frontend then fetches the public key from the backend canister. A possible yet minor optimization would be to directly derive this key in the frontend (which can be found in the developer documentation), instead of making canister calls

since this also saves the latency for the canister call.

randombit (approved):

examples/basic_timelock_ibe/README.md
fspreiss: Hm...the latency should be irrelevant, no matter if the key is derived in the frontend or the backend, because in both cases there is no canister call needed.

altkdf: Currently we fetch the root IBE key for the frontend from the backend and the backend fetches it from the management canister, right? If we hard-code the subnet key in the frontend, we can derive the root IBE key for the dapp without any interaction with the IC. Or are we talking about different things?

fspreiss: > Currently we fetch the root IBE key for the frontend from the backend and the backend fetches it from the management canister, right?

Right!

> If we hard-code the subnet key in the frontend, we can derive the root IBE key for the dapp without any interaction with the IC.

Correct, and what I'm trying to say is that even if we hard-code the subnet key in the _backend_ instead of the frontend and derive the IBE key in the canister code (which is what my suggested paragraph above says), then we _also_ don't need any interaction with the IC, i.e., latency-wise, this is equivalent.

So, what we could do is adapt your sentence to: "A possible yet minor optimization would be to directly derive this key in the frontend or the backend..."

altkdf: replaced with
> **Public Key Retrieval**: It is possible to use one public key to encrypt bids for multiple auction lots. The subkey derivation for a concrete lot can happen on the client side, i.e., in the frontend. It is already the case in this example. A further possible yet minor optimization would be to directly derive this key in the frontend or the backend instead of fetching it from the management canister, thus reducing the latency.

I guess what I meant with "latency" was the network communication latency of the frontend, which affects the UX, although only slightly, if the user has a good internet connection. I was not strictly talking about the consensus latency of the update calls or inter-canister calls. The network round trip is still there for a query call. But I guess we are making this small remark much more important than it is by discussing it so extensively.

fspreiss: Let's discuss it next time we are in a Zoom call :-)

PR #88, opened 2025-05-07
chore: CRP-2768 use shared `dfx` network in frontend tests

altkdf:
Remark: port 4943 is the default port for the shared network. Port 8000 is the commonly used port for the local network (i.e., `dfx` needs to run from within the project folder). The only case where we have dependency on multiple projects is the frontend tests, where we run all tests in one go. Therefore, we need the key manager and encrypted maps tests to both be available.

So far we used two different ports for that, but this is somewhat confusing. This PR changes the `dfx` config to use a shared network with the same port (4943) for the canisters used in the tests.

### DIFF
--- a/examples/basic_timelock_ibe/backend/src/lib.rs
+++ b/examples/basic_timelock_ibe/backend/src/lib.rs
@@ -33,7 +33,7 @@ thread_local! {
 
     static VETKD_ROOT_IBE_PUBLIC_KEY: RefCell<Option<VetKeyPublicKey>> =  const { RefCell::new(None) };
 
-    static BID_COUNTER: RefCell<BidCounter> = RefCell::new(0);
+    static BID_COUNTER: RefCell<BidCounter> = const { RefCell::new(0) };
 
     #[cfg(feature = "expose-testing-api")]
     static CANISTER_ID_VETKD_MOCK: RefCell<Option<Principal>> = const { RefCell::new(None) };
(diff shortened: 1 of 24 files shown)


altkdf:
This PR also fixes a `clippy` problem that was previously uncovered because clippy runs only on changes to the backend. I added a workflow in https://github.com/dfinity/vetkd-devkit/pull/92. That PR will be rebased once this goes in to fix the failing CI.

altkdf:
P.S.: it doesn't seem easy to change the port of the shared setting because the config is stored somewhere under `$HOME`. So this feels to me like a slightly cleaner solution, although 2 different ports.

examples/basic_timelock_ibe/backend/src/lib.rs
```diff
@@ -178,8 +178,7 @@ fn place_bid(lot_id: u128, encrypted_amount: Vec<u8>) -> Result<(), String> {
         if let Some((existing_bid_key, _existing_bid)) = bids
             .range((lot_id, 0, Principal::management_canister())..)
             .take_while(|((this_lot_id, _, _), _)| *this_lot_id == lot_id)
-            .filter(|((_, _, this_bidder), _)| *this_bidder == bidder)
-            .next()
+            .find(|((_, _, this_bidder), _)| *this_bidder == bidder)
```
fspreiss, line 181: Is it intentional that the changes in this file are part of the PR?

fspreiss (approved):

PR #90, opened 2025-05-07
chore(crypto): CRP-2789 Use a fixed length domain separator for MaskMsg

randombit:


frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -493,7 +493,7 @@ function maskSeed(seed: Uint8Array, t: Uint8Array): Uint8Array {
 }
 
 function maskMsg(msg: Uint8Array, seed: Uint8Array): Uint8Array {
-    const domain_sep = IbeDomainSeparators.MaskMsg.concat(msg.length.toString());
+    const domain_sep = IbeDomainSeparators.MaskMsg.concat(msg.length.toString().padStart(20, "0"));
```
altkdf, line 496: copy-paste the Rust comment here? https://github.com/dfinity/ic/pull/5028/files#diff-5aa6fc7223c2b61a027e528ec4c27df2f22296595d6534bd59df4f554c111177R372

altkdf: Should we also bump `ic_vetkd_utils` s.t. the domain separators in `basic_timelock_ibe` are consistent across languages, since there the frontend encrypts and the backend decrpyts IBE ciphertexts? I think it matters in the decryption, right?

randombit: Comment introduced, also bumped `Cargo.lock` to the matching revision of the monorepo

PR #92, opened 2025-05-08
ci: add a clippy workflow that also runs for backends in the examples

altkdf:


altkdf:
Merging `main` after https://github.com/dfinity/vetkd-devkit/pull/88 goes in should fix the CI errors.

.github/workflows/clippy.yml
```diff
@@ -0,0 +1,41 @@
...
+      - examples/**/backend/**
+      - Cargo.toml
+      - Cargo.lock
+      - .github/workflows/provision-linux.sh
+      - .github/workflows/provision-darwin.sh
+      - .github/workflows/clippy.yml
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+jobs:
```
fspreiss, line 18: When we created the existing clippy job, we discussed if clippy needs to be run for both linux and macos, and concluded one should be enough.

Here we run it for both linux and macos. I'm OK with that, but am wondering if there were new insights.

fspreiss (approved):

.github/workflows/clippy.yml
altkdf: I guess it still doesn't *need* to run it. This is rather a sanity check. We could also remove it if you think that not having it makes more sense.

altkdf: removed

PR #95, opened 2025-05-13
chore: adapt frontend library `package.json` to be better suitable for publishing

altkdf:
- Changed the name of the package to `@dfinity/vetkeys`. (Having `ic_` prefix would be unexpected given that no other package in the `@dfinity` scope has such a prefix and they all are about IC.)
- Added some metadata like keywords, URLs, etc. to `package.json`.
- As a side-effect of looking at it, fixed the `coverage` script, since it requires canister deployment, and added an ignore to eslint for coverage.

### DIFF
--- a/frontend/ic_vetkeys/package.json
+++ b/frontend/ic_vetkeys/package.json
@@ -1,7 +1,42 @@
 {
-    "name": "ic_vetkeys",
+    "name": "@dfinity/vetkeys",
     "version": "0.1.0",
+    "author": "DFINITY Stiftung <sdk@dfinity.org>",
+    "description": "JavaScript and TypeScript library to interact with vetKeys on the Internet Computer",
+    "homepage": "https://internetcomputer.org/docs/building-apps/network-features/encryption/vetkeys",
     "license": "Apache-2.0",
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/dfinity/vetkd-devkit.git",
+        "directory": "frontend/ic_vetkeys"
+    },
+    "bugs": {
+        "url": "https://github.com/dfinity/vetkd-devkit/issues"
+    },
+    "keywords": [
+        "internet computer",
+        "internet-computer",
+        "ic",
+        "icp",
+        "dfinity",
+        "vetkeys",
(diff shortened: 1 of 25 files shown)


altkdf:
I used the [`package.json` of `@dfinity/agent`](https://github.com/dfinity/agent-js/blob/e7332d45dadab979f32efef5d11874b262321229/packages/agent/package.json) as inspiration.

altkdf:
What should be the contact email, if any? Currently I just added `sdk@dfinity.org` as a placeholder.

frontend/ic_vetkeys/package.json
```diff
@@ -1,7 +1,42 @@
 {
-    "name": "ic_vetkeys",
+    "name": "@dfinity/vetkeys",
     "version": "0.1.0",
+    "author": "DFINITY Stiftung <sdk@dfinity.org>",
+    "description": "JavaScript and TypeScript library to interact with vetKeys on the Internet Computer",
```
fspreiss, line 5: How about "...to _use_ vetKeys..."?

fspreiss:
> What should be the contact email, if any? Currently I just added sdk@dfinity.org as a placeholder.

Given that our team doesn't really have an email address, I'd just leave it out (at least for now). People can always create issues on Github.

fspreiss (approved):
Thanks, @altkdf! Before we publish, maybe we can quickly discuss the naming: currently the code lives in the "tools" folder. And there is also a "utils" folder. The two seem somewhat interchangable, and it would be good if we make a conscious decision on the team regarding this.

frontend/ic_vetkeys/package.json
altkdf: Changed to `JavaScript and TypeScript library to use Internet Computer vetKeys`.

altkdf:
> Thanks, @altkdf! Before we publish, maybe we can quickly discuss the naming: currently the code lives in the "tools" folder. And there is also a "utils" folder. The two seem somewhat interchangable, and it would be good if we make a conscious decision on the team regarding this.

This will be refactored in another PR.

PR #96, opened 2025-05-14
chore: fix ci by fixing two failing tests and add a workflow dependency

altkdf:
The workflow dependency is due to the frontend tests being dependent on backend canisters.

### DIFF
--- a/frontend/ic_vetkeys/src/encrypted_maps/encrypted_maps_canister.test.ts
+++ b/frontend/ic_vetkeys/src/encrypted_maps/encrypted_maps_canister.test.ts
@@ -223,7 +223,7 @@ test("can get user rights", async () => {
     const user = id1.getPrincipal();
     const encryptedMapsOwner = await newEncryptedMaps(id0);
     const encryptedMapsUser = await newEncryptedMaps(id1);
-    const rights = { ReadWrite: null };
+    const rights = { ReadWriteManage: null };
 
     await encryptedMapsOwner.setValue(
         owner,
--- a/frontend/ic_vetkeys/src/key_manager/key_manager_canister.test.ts
+++ b/frontend/ic_vetkeys/src/key_manager/key_manager_canister.test.ts
@@ -116,7 +116,7 @@ test("sharing rights are consistent", async () => {
     const keyManagerUser = await newKeyManager(id1).catch((err) => {
         throw err;
     });
-    const rights = { ReadWrite: null };
+    const rights = { ReadWriteManage: null };
 
     expect(
         await keyManagerOwner.setUserRights(
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,10 +9,7 @@
                 "examples/basic_ibe/frontend",
                 "examples/password_manager/frontend",
                 "examples/password_manager_with_metadata/frontend"
-            ],
-            "peerDependencies": {
-                "@dfinity/vetkeys": "^0.1.0"
-            }
+            ]
         },
         "examples/basic_ibe/frontend": {
             "version": "0.0.0",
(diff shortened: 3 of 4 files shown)


package-lock.json
```diff
@@ -9,10 +9,7 @@
                 "examples/basic_ibe/frontend",
                 "examples/password_manager/frontend",
                 "examples/password_manager_with_metadata/frontend"
-            ],
-            "peerDependencies": {
-                "@dfinity/vetkeys": "^0.1.0"
```
altkdf, line 14: not sure why this was there, but it seems something automatically removed it

frontend/ic_vetkeys/src/encrypted_maps/encrypted_maps_canister.test.ts
```diff
@@ -223,7 +223,7 @@ test("can get user rights", async () => {
     const user = id1.getPrincipal();
     const encryptedMapsOwner = await newEncryptedMaps(id0);
     const encryptedMapsUser = await newEncryptedMaps(id1);
-    const rights = { ReadWrite: null };
+    const rights = { ReadWriteManage: null };
```
fspreiss, line 226: Somewhat unrelated, but I'm confused by the test: in lines 256 to 261: `owner` is the caller and asking for the map `encryptedMapsUser` the user-rights of key `some key` for user `user`, and that returns `ReadWriteManage`: but is `owner` really allowed to ask for that? It is not "his" map, but he can still see the rights...maybe I'm missing something.

altkdf: Hmm, `encryptedMaps*` objects could also be though of as `encryptedMapsClient*` objects. So they just connect to the same canister with different principals. What happens in the test is the following:
1. `owner` asks for permissions of `user` in `(owner, "some map")`, gets `undefined`
2. `owner` adds permissions for `user` in ` (owner, "some map")`, gets back previous permissions `undefined`
3. `owner` asks again for permissions, now gets `rights`.
4. `user` asks for permissions of `user` in`(owner, "some map")`, gets `rights`.

frontend/ic_vetkeys/src/key_manager/key_manager_canister.test.ts
```diff
@@ -116,7 +116,7 @@ test("sharing rights are consistent", async () => {
     const keyManagerUser = await newKeyManager(id1).catch((err) => {
         throw err;
     });
-    const rights = { ReadWrite: null };
+    const rights = { ReadWriteManage: null };
```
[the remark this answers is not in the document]
altkdf: Yes because there are 2 key manager clients: `keyManagerOwner` and `keyManagerUser`. For the owner, it doesn't matter what permissions the client has. But at the end of the test, we are also using the latter to essentially check user's own permissions. And any user can only get anyone's access permissions with `ReadWriteManage` own permissions.

`
expect(
        await keyManagerUser.getUserRights(
            owner,
            new TextEncoder().encode("some key"),
            user,
        ),
    ).to.deep.equal(rights);
`
reads as `user` (because `keyManagerUser`) gets permissions of `user` for key `(owner, b"some_key")`.

frontend/ic_vetkeys/src/encrypted_maps/encrypted_maps_canister.test.ts
altkdf: I would say let's merge this now because https://github.com/dfinity/vetkd-devkit/pull/97 depends on it, and let's discuss this separately

frontend/ic_vetkeys/src/key_manager/key_manager_canister.test.ts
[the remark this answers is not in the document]
fspreiss: Ah, OK! Now I understand, thanks! I wrongly thought that the first parameter is the caller. Sorry for the noise!

PR #97, opened 2025-05-14
chore(crypto): CRP-2794 Modify VetKD key derivation scheme

randombit:


altkdf:
Yes: https://github.com/dfinity/vetkd-devkit/pull/96

andreacerulli (approved):
Key derivation changes LGTM

PR #101, opened 2025-05-19
chore: CRP-2742 better linting in IBE examples and small fixes across all examples

altkdf:


examples/basic_ibe/backend/src/lib.rs
```diff
@@ -84,10 +84,11 @@ async fn get_my_encrypted_ibe_key(transport_key: TransportPublicKey) -> Encrypte
         transport_public_key: transport_key.into_vec(),
     };
 
-    let (result,) = ic_cdk::api::call::call::<_, (VetKDDeriveKeyReply,)>(
+    let (result,) = ic_cdk::api::call::call_with_payment128::<_, (VetKDDeriveKeyReply,)>(
         vetkd_system_api_canister_id(),
         "vetkd_derive_key",
         (request,),
+        26_153_846_153,
```
altkdf, line 91: this now fails because we are using dfx and not chainkey testing canister

examples/password_manager/frontend/src/lib/encrypted_maps.ts
```diff
@@ -11,7 +11,7 @@ export async function createEncryptedMaps(
     const host =
         process.env.DFX_NETWORK === "ic"
             ? `https://${process.env.CANISTER_ID_IC_VETKEYS_ENCRYPTED_MAPS_CANISTER}.ic0.app`
-            : "http://localhost:8000";
+            : "http://localhost:4943";
```
altkdf, line 14: port has changed (4943 is a shared replica, we need it now for running tests in the frontend library)

examples/password_manager/frontend/src/store/vaults.ts
```diff
@@ -126,9 +126,9 @@ export async function removeUser(
     );
 }
 
-auth.subscribe(() => {
-    void (async ($auth) => {
-        if ($auth && $auth.state === "initialized") {
+auth.subscribe((auth) => {
```
altkdf, line 129: this was sometimes working, but often failing, this changes seems to fix the problem

examples/basic_ibe/deploy_locally.sh
```diff
@@ -13,9 +13,11 @@ dfx ping &> /dev/null || dfx start --background --clean >> /dev/null
 dfx deps pull && dfx deps init && dfx deps deploy &&
     export CANISTER_ID_INTERNET_IDENTITY=rdmx6-jaaaa-aaaaa-aaadq-cai
 
+dfx canister create basic_ibe
+
 # Store environment variables for the frontend.
 echo "DFX_NETWORK=$DFX_NETWORK" > frontend/.env
-echo "CANISTER_ID_BASIC_IBE=$CANISTER_ID_BASIC_IBE" >> frontend/.env
+echo "CANISTER_ID_BASIC_IBE=$(dfx canister id basic_ibe)" >> frontend/.env
```
altkdf, line 20: Here, it is more of a question why this worked before before explicitly defining the canister id in the shell. I was initially assuming that `dfx` somehow exports environment variables, but it doesn't seem to be true. It now fails without the change (the frontend canister is built, but the backend canister id is not set and calls to the backend canister from the user frontend fail)

altkdf:
Closes CRP-2742.

examples/basic_ibe/frontend/eslint.config.mjs
```diff
@@ -1,10 +1,31 @@
...
+import tseslint from "typescript-eslint";
+import eslintPluginPrettierRecommended from "eslint-plugin-prettier/recommended";
 
 export default tseslint.config(
-  eslint.configs.recommended,
-  tseslint.configs.recommended,
-  { ignores: ["dist/", "babel.config.js", "src/declarations"] }
+    eslint.configs.recommended,
+    tseslint.configs.recommended,
+    tseslint.configs.recommendedTypeChecked,
```
fspreiss, line 10: Are both lines needed? Looking at [this](https://typescript-eslint.io/getting-started/typed-linting), it seems only `tseslint.configs.recommendedTypeChecked` is needed...

examples/basic_timelock_ibe/frontend/eslint.config.mjs
```diff
@@ -1,10 +1,31 @@
...
+import tseslint from "typescript-eslint";
+import eslintPluginPrettierRecommended from "eslint-plugin-prettier/recommended";
 
 export default tseslint.config(
-  eslint.configs.recommended,
-  tseslint.configs.recommended,
-  { ignores: ["dist/", "babel.config.js", "src/declarations"] }
+    eslint.configs.recommended,
+    tseslint.configs.recommended,
+    tseslint.configs.recommendedTypeChecked,
```
fspreiss, line 10: Same question as before: are both needed?

fspreiss (approved):

examples/basic_ibe/frontend/eslint.config.mjs
altkdf: You're right, `recommendedTypeChecked` contains `recommended`. Removed  everywhere.

PR #103, opened 2025-05-21
feat: Export a function for verifying BLS signatures from TS utils

randombit:
Makes life a little easier for end users who might want to for example verify a VRF output

### DIFF
--- a/frontend/ic_vetkeys/src/utils/utils.ts
+++ b/frontend/ic_vetkeys/src/utils/utils.ts
@@ -320,14 +320,16 @@ export function augmentedHashToG1(
 export function verifyBlsSignature(
     pk: DerivedPublicKey,
     message: Uint8Array,
-    signature: G1Point,
+    signature: G1Point | Uint8Array,
 ): boolean {
     const neg_g2 = bls12_381.G2.ProjectivePoint.BASE.negate();
     const gt_one = bls12_381.fields.Fp12.ONE;
 
+    const signaturePt = signature instanceof bls12_381.G1.ProjectivePoint ? signature : bls12_381.G1.ProjectivePoint.fromHex(signature);
+
     const messageG1 = augmentedHashToG1(pk, message);
     const check = bls12_381.pairingBatch([
-        { g1: signature, g2: neg_g2 },
+        { g1: signaturePt, g2: neg_g2 },
         { g1: messageG1, g2: pk.getPoint() },
     ]);
 
--- a/frontend/ic_vetkeys/src/utils/utils.ts
+++ b/frontend/ic_vetkeys/src/utils/utils.ts
@@ -326,7 +326,10 @@ export function verifyBlsSignature(
     const neg_g2 = bls12_381.G2.ProjectivePoint.BASE.negate();
     const gt_one = bls12_381.fields.Fp12.ONE;
 
-    const signaturePt = signature instanceof bls12_381.G1.ProjectivePoint ? signature : bls12_381.G1.ProjectivePoint.fromHex(signature);
+    const signaturePt =
+        signature instanceof bls12_381.G1.ProjectivePoint
+            ? signature
+            : bls12_381.G1.ProjectivePoint.fromHex(signature);
 
     const messageG1 = augmentedHashToG1(pk, message);
     const check = bls12_381.pairingBatch([
--- a/frontend/ic_vetkeys/src/utils/utils.ts
+++ b/frontend/ic_vetkeys/src/utils/utils.ts
@@ -307,6 +307,34 @@ export function augmentedHashToG1(
 }
 
 /**
+ * Verify a BLS signature
+ *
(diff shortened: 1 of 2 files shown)


frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -306,6 +306,34 @@ export function augmentedHashToG1(
...
+ * message.
+ *
+ * When a VetKey struct is created (VetKey.decryptAndVerify) the signature (aka
+ * vetkey) is already verified, so using this function is only necessary when
+ * using a vetkey as a VRF or for threshold BLS signatures.
+ */
+export function verifyBlsSignature(
+    pk: DerivedPublicKey,
+    message: Uint8Array,
+    signature: G1Point,
```
altkdf, line 323: Can we do here `signature: G1Point | Uint8Array` or show an example in the docs how do convert bytes to `G1Point`?

altkdf:
We will also need a similar function in Rust

frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -306,6 +306,34 @@ export function augmentedHashToG1(
...
 }
 
+/**
+ * Verify a BLS signature
+ *
+ * A VetKey is in the end a valid BLS signature; this function checks that a
+ * provided BLS signature is the valid one for the provided public key and
+ * message.
+ *
+ * When a VetKey struct is created (VetKey.decryptAndVerify) the signature (aka
```
fspreiss, line 316: IIRC, `decryptAndVerify` lives on `EncryptedVetKey` (not `VetKey`)

frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -306,6 +306,34 @@ export function augmentedHashToG1(
...
+ * vetkey) is already verified, so using this function is only necessary when
```
fspreiss, line 317: maybe add ", which can be obtained via `VetKey::signatureBytes`, "

nit: maybe also write "vetKey" instead of "vetkey" whenever we are not talking about the TypeScript class VetKey (or Rust struct VetKey).

frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -566,22 +594,14 @@ export class EncryptedVetKey {
             throw new Error("Invalid VetKey");
         }
 
-        // Compute the purported vetkd k
+        // Compute the purported vetkey k
```
fspreiss, line 597:
```suggestion
        // Compute the purported vetKey k
```

frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -306,6 +306,34 @@ export function augmentedHashToG1(
...
+ *
+ * A VetKey is in the end a valid BLS signature; this function checks that a
+ * provided BLS signature is the valid one for the provided public key and
+ * message.
+ *
+ * When a VetKey struct is created (VetKey.decryptAndVerify) the signature (aka
+ * vetkey) is already verified, so using this function is only necessary when
+ * using a vetkey as a VRF or for threshold BLS signatures.
+ */
+export function verifyBlsSignature(
```
fspreiss, line 320: Should we add dedicated tests for this new public function, or are the indirect tests we already have sufficient?

fspreiss (commented):
Nice, thanks @randombit!

fspreiss:
> We will also need a similar function in Rust

Indeed, we should also have this for Rust in the `ic_vetkeys` crate. I wonder if underneath this should call out to our [ic-verify-bls-signature](https://crates.io/crates/ic-verify-bls-signature). The latter would have to be adapted to support the augmented scheme though.

frontend/ic_vetkeys/src/utils/utils.ts
randombit: Fixed thanks

frontend/ic_vetkeys/src/utils/utils.ts
randombit: Fixed

frontend/ic_vetkeys/src/utils/utils.ts
randombit: Fixed

frontend/ic_vetkeys/src/utils/utils.ts
randombit: IMO the existing usage in decryptAndVerify was sufficient for how it was, but did not cover the case for where we passed an array of bytes instead of a point. I added a test.

frontend/ic_vetkeys/src/utils/utils.ts
randombit: Extended accordingly, with a test that everything works as expected

altkdf (approved):

fspreiss (approved):
Thanks, @randombit!

PR #105, opened 2025-05-22
docs: CRP-2676 improve rust backend docs

altkdf:


README.md
```diff
@@ -1,51 +1,52 @@
...
 
-**VetKeys** – Verifiable Encrypted Threshold Keys – on the Internet Computer addresses the fundamental challenge of storing secrets on-chain by allowing cryptographic key derivation without exposing private keys to anyone but the user. By leveraging **threshold cryptography**, VetKeys make it possible to generate, transport, and use encrypted keys securely, unlocking **privacy-preserving smart contracts** and **externally verifiable randomness**.
+**vetKeys** – Verifiable Encrypted Threshold Keys – on the Internet Computer addresses the fundamental challenge of storing secrets on-chain by allowing cryptographic key derivation without exposing private keys to anyone but the user. By leveraging **threshold cryptography**, vetKeys make it possible to generate, transport, and use encrypted keys securely, unlocking **privacy-preserving smart contracts** and **externally verifiable randomness**.
 
-In slightly more detail, VetKeys enables use cases such as:
+In slightly more detail, vetKeys enables use cases such as:
```
fspreiss, line 12: nit: I personally would write, e.g., "the vetKeys feature" (or similar) whenever we mean the *feature*, and otherwise, when "vetKeys" stands by itself, let it mean the plural of a single vetKey.

This because reading "vetKeys enables" seems weird to me.

So, here in this line, we could write
* "In slightly more detail, vetKeys enable use cases such as:"
* "In slightly more detail, the vetKeys feature enables use cases such as:"

In the entire PR here, this was the only occurrence that I'd change, all other usages of vetKeys sounded good to me. I'm happy to create a PR to fix this one occurrence if we agree that this makes sense.

fspreiss (commented):
Thank, @altkdf! Also looks good to me. Sorry for the late review.

README.md
altkdf: No strong opinion from my side, feel free to change :+1:

PR #108, opened 2025-05-22
feat: Add BLS signature verification function to Rust utils

randombit:


altkdf (approved):

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -589,6 +586,36 @@ impl IBECiphertext {
     }
 }
 
+/// Verify an augmented BLS signature
+///
+/// Returns true if and only if the provided signature is valid with respect to
+/// the provided public key and input
+pub fn verify_bls_signature(dpk: &DerivedPublicKey, input: &[u8], signature: &[u8]) -> bool {
+    let signature: G1Affine = match signature.try_into() {
```
fspreiss, line 594: nit: slight preference for being explicit about what we are trying to convert into:
```
let signature = match <[u8; 48]>::try_from(signature) {
```

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -589,6 +586,36 @@ impl IBECiphertext {
     }
 }
 
+/// Verify an augmented BLS signature
```
fspreiss, line 589: It might make sense to give more details on what "augmented" means. So maybe turn "augmented" into a link to
* https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature#name-message-augmentation-2 or
* https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature#name-message-augmentation

fspreiss (approved):
Thanks, @randombit!

backend/rs/ic_vetkeys/src/utils/mod.rs
randombit: Done

backend/rs/ic_vetkeys/src/utils/mod.rs
randombit: Done

PR #111, opened 2025-05-23
feat: CRP-2818 Add types for IBE identity and seed parameters

randombit:


backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -397,14 +397,98 @@ impl EncryptedVetKey {
...
+///
+/// As far as the IBE encryption scheme goes this is simply an opauqe bytestring
+/// We provide a type to make code using the IBE a bit easier to understand
+pub struct Identity {
+    val: Vec<u8>,
+}
+
+impl Identity {
+    /// Create an identity from an arbitrary byte string
+    pub fn from_bytes(bytes: &[u8]) -> Self {
```
altkdf, line 411: Why not use Rust's `From`?

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -397,14 +397,98 @@ impl EncryptedVetKey {
     }
 }
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+/// An identity, used for identity based encryption
```
altkdf, line 401:
```suggestion
/// An identity, used for identity based encryption (IBE)
```

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -397,14 +397,98 @@ impl EncryptedVetKey {
...
+///
+/// As far as the IBE encryption scheme goes this is simply an opauqe bytestring
```
altkdf, line 403:
```suggestion
/// As far as the IBE scheme goes this is simply an opauqe bytestring
```

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -397,14 +397,98 @@ impl EncryptedVetKey {
...
+    }
+}
+
+/*
+* Amount of randomness generated during the IBE encryption process
+*/
 const IBE_SEED_BYTES: usize = 32;
 
+/// A random seed, used for identity based encryption
+pub struct Seed {
```
altkdf, line 439: It might be better to put this into an `ibe`/`identity_based_encryption` module, then it would be very clear what context `Identity` and `Seed` have.

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -397,14 +397,98 @@ impl EncryptedVetKey {
...
+    }
+
+    /// Create a seed for IBE encryption from a byte string
+    ///
+    /// This input should be randomly chosen by a secure random number generator.
+    /// If the seed is not securely generated the IBE scheme will be insecure.
+    ///
+    /// If the input is exactly 256 bits it is used directly. Otherwise the input
+    /// is hashed with HKDF to produce a 256 bit seed.
+    pub fn from_bytes(bytes: &[u8]) -> Self {
```
altkdf, line 459: `from_seed`?

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -397,14 +397,98 @@ impl EncryptedVetKey {
...
+            val: rng.gen::<[u8; IBE_SEED_BYTES]>(),
+        }
+    }
+
+    /// Create a seed for IBE encryption from a byte string
+    ///
+    /// This input should be randomly chosen by a secure random number generator.
+    /// If the seed is not securely generated the IBE scheme will be insecure.
+    ///
+    /// If the input is exactly 256 bits it is used directly. Otherwise the input
```
altkdf, line 457: Do we want to return an error if the input is too small?

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -397,14 +397,98 @@ impl EncryptedVetKey {
...
+ * IBE ciphertexts are prefixed with a header to identity the protocol and provide
+ * an extension point if needed in the future eg for changing to a different cipher.
+ *
+ * The header consists of "IC IBE" (ASCII) plus two bytes 0x00 and 0x01 which
+ * here are just fixed and effectively arbitrary values, but could be used to
+ * indicate for example a version in the future should we need to support multiple
+ * variants of the IBE scheme.
+*/
 const IBE_HEADER_BYTES: usize = 8;
 const IBE_HEADER: [u8; IBE_HEADER_BYTES] = [b'I', b'C', b' ', b'I', b'B', b'E', 0x00, 0x01];
```
altkdf, line 487: Maybe `IBE_HEADER` would be enough, then `IBE_HEADER_BYTES=IBE_HEADER.len()`. And then the rustdoc would be on `IBE_HEADER.

frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -711,12 +712,109 @@ function isEqual(x: Uint8Array, y: Uint8Array): boolean {
     return diff == 0;
 }
 
+/**
+ * An identity used for identity based encryption
+ *
+ * As far as the IBE encryption scheme goes this is simply an opauqe bytestring
+ * We provide a type to make code using the IBE a bit easier to understand
+ */
+export class Identity {
```
altkdf, line 721: Ah, we want to have a very similar interface for both. That's already an improvement. So if neither you @randombit nor @fspreiss has ideas how to structure it in a way to make it more explicit that `Identity` and `Seed` are about IBE, maybe we will want to merge it with the current names.

backend/rs/ic_vetkeys/src/utils/mod.rs
fspreiss: It's indeed an option. But having `from_bytes` next to `from_string` and `from_principal` is also quite nice because then all conversions are explicit and next to each other, rather than one type of conversion being a bit harder to spot in the API docs. I'm fine either way.

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -535,29 +619,20 @@ impl IBECiphertext {
...
     /// and with an `input` equal to the provided `identity` parameter.
-    pub fn encrypt(
-        dpk: &DerivedPublicKey,
-        identity: &[u8],
-        msg: &[u8],
-        seed: &[u8],
-    ) -> Result<IBECiphertext, String> {
-        let seed: &[u8; IBE_SEED_BYTES] = seed
-            .try_into()
-            .map_err(|_e| format!("Provided seed must be {} bytes long ", IBE_SEED_BYTES))?;
```
fspreiss, line 546: We can remove the respective comment ("seed must be exactly 256 bits") in the docs.

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -434,7 +518,7 @@ impl IBEDomainSep {
     }
 }
 
-impl IBECiphertext {
+impl IbeCiphertext {
```
fspreiss, line 521: Maybe also rename `IBEDomainSep` to `IbeDomainSep` and/or move it to a module for IBE.

backend/rs/ic_vetkeys/src/utils/mod.rs
fspreiss: A module for the IBE types (e.g., called `identity_based_encryption` or just `ibe`) might indeed be nice.

If we don't move the types to such a module, and we don't intend to use `Seed` and `Identity` for other purposes than IBE (and I believe we don't) then the names `IbeIdentity` and `IbeSeed` would also work.

I'd be fine either way.

frontend/ic_vetkeys/src/utils/utils.ts
fspreiss: Can we have modules in Typescript? If not, then all I can think of is to keep the names, or use `IbeIdentity` and `IbeSeed`. No strong preference.

backend/rs/ic_vetkeys/src/utils/mod.rs
randombit: Changed

backend/rs/ic_vetkeys/src/utils/mod.rs
randombit: Renamed to `IbeDomainSep` to match `IbeCiphertext`

backend/rs/ic_vetkeys/src/utils/mod.rs
randombit: Fixed

backend/rs/ic_vetkeys/src/utils/mod.rs
randombit: Trying to keep TS and Rust in sync I just went with `IbeIdentity` and `IbeSeed` for both languages

backend/rs/ic_vetkeys/src/utils/mod.rs
altkdf: `from_seed_bytes`? :) compare e.g. with `Identity::from_bytes`. I intuitively think of something like that.

frontend/ic_vetkeys/src/utils/utils.ts
altkdf: There is definitely a way. But I'm not sure if that will look better, so probably let's use `Ibe*` names for now and think about if we can improve it when we have time?

altkdf (approved):

backend/rs/ic_vetkeys/src/utils/mod.rs
randombit: > I guess that's useful it you instantiate something like
> KeyManager<I: Into<Identity>> {...}

OK yeah that's a good point. We can always add `From` conversions later if it seems helpful.

backend/rs/ic_vetkeys/src/utils/mod.rs
randombit: I had wanted to make it infalliable but ignoring eg empty input is probably a footgun. I changed it to require at least 128 bits since that ~~ the security level of BLS12-381 anyway. Hopefully everyone will just use `IbeSeed::random` instead

backend/rs/ic_vetkeys/src/utils/mod.rs
randombit: Going to leave it like this for now, we can always iterate if it seems devs are getting confused about how to use the API

PR #112, opened 2025-05-23
chore: CRP-2816 add key initialization

altkdf:
So far we always initialized canisters with a hardcoded `dfx_test_key`. This change adds an `init` routine that stores the key. This now makes sense to do, since soon we will want to test the examples on mainnet.


altkdf:
I manually tested the canisters and so far everything I tested (hopefully everything) worked.

randombit (approved):

PR #115, opened 2025-05-26
chore: CRP-2796 update Cargo.toml for publishing

altkdf:


altkdf:
Took a few things from https://github.com/dfinity/cdk-rs/blob/main/Cargo.toml.

fspreiss (approved):
Thanks, @altkdf!

PR #116, opened 2025-05-26
chore: CRP-2795 CRP-2796 add workflows for publishing crates and packages

altkdf:
Adds 3 workflows:
* Publish frontend package. Creates tag and release from a PR branch of format `release/frontend/@dfinity/vetkeys/vx.y.z`, this will create a tag without the `release/` prefix. Checks that the version has valid format and that it is the same version as in package.json. 
* Publish docs for the frontend package on github.io. Works via a manual workflow trigger on a tag. This is s.t. we can also add backports and would not need to overwrite the docs because we only publish the latest version of the docs.
* Publish backend crate *release*. Does the same as the frontend workflow, but does not publish the crate on crates.io. That would be a manual process. The format of the branch is `release/backend/ic_vetkeys/vx.y.z`.

Also fixes that in the frontend we include testing artifacts in the build.


fspreiss (approved):
Thanks, @altkdf!

PR #124, opened 2025-05-27
ci: CRP-2795 add write permissions for `id-token` that are required to use `npm publish --provenance`

altkdf:


altkdf:
ChatGPT explain this dependency as follows:
>When you run npm publish --provenance in GitHub Actions, the npm CLI under the hood uses Sigstore’s OpenID Connect (OIDC)–based flow to mint a signed provenance attestation. In order to do that, your workflow needs to request an OIDC JWT from GitHub’s identity provider—and GitHub will only hand out that token if you’ve explicitly granted your job write access to the id-token permission.

Running `npm publish` with  `--provenance` seems to be a good security practice, so it probably makes sense to keep it.

PR #129, opened 2025-05-30
chore: check formatting using `mops format --check` in a workflow.

altkdf:
Similar to `cargo fmt --check`.


altkdf:
Since [recently](https://github.com/ZenVoich/mops/issues/280#issuecomment-2915905800), `mops` (starting from `v1.8.0`) supports formatting and format checking. This makes it quite easy to use it in this repo.

PR #133, opened 2025-06-03
chore: in examples fix incorrect handling of some key ids and mainnet frontend configuration

altkdf:
- Fixed Problem 1: In some cases `agentOptions` were not passed down to the actor creation function if deploying the frontend to mainnet.

- Fixed Problem 2: In some cases the key name that was set on canister initialization was ignored and instead `"dfx_test_key"` was always used.


altkdf:
Change the PR title to chore instead of fix because the change only affects examples, which are not published anywhere expect github.

fspreiss (approved):

PR #134, opened 2025-06-04
chore: CRP-2834 use `dfx deploy` to simplify setup

altkdf:


fspreiss (approved):

altkdf:
@fspreiss I manually deployed the examples and everything worked so far. But at some point we should double-check everything and make sure it also deploys on mainnet without problems.

altkdf:
I made a few more modifications, so you may want to have another look @fspreiss .

fspreiss (approved):
Still looks good, thanks, @altkdf!

PR #135, opened 2025-06-04
chore: CRP-2823 modularize parts of the vetKD-related code in examples to simplify them

altkdf:
This makes it easier to use them in developer docs.


altkdf:
Does this make sense to you @fspreiss ?

altkdf:
Note: I didn't yet check that nothing is broken in the examples.

PR #140, opened 2025-06-06
chore: CRP-2829 release Rust backend `v0.2.0`

altkdf:


backend/rs/ic_vetkeys/CHANGELOG.md
```diff
@@ -1,10 +1,16 @@
 # Change Log
 
-## Unreleased
+## [0.2.0] - 06.06.2025
```
fspreiss, line 3: The date needs to be bumped...

fspreiss (approved):

backend/rs/ic_vetkeys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ic-vetkeys"
-version = "0.1.0"
+version = "0.2.0"
```
fspreiss, line 3: Not sure if we will be able to always keep the TS and Rust version in sync, as one may need a bug fix that the other one doesn't. Anyway, we should try.

altkdf: Yes, maybe in the future we could use major versions for that.

PR #142, opened 2025-06-09
chore: CRP-2718 Add function for checking validity of transport public keys

randombit:


frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -68,6 +68,26 @@ export class TransportSecretKey {
     }
 }
 
+/**
+ * Check if a transport public key is valid
+ *
+ * This tests if the passed byte array is of the expected size and encodes
+ * a valid group element.
+ */
+export function isValidTransportPublicKey(tpk: Uint8Array): boolean {
```
fspreiss, line 77: Note that in the Rust backend utils we call it `is_valid_transport_public_key_encoding` (note the "encoding" part), but I actually like this name here (without encoding) a bit better. I guess we can live with the fact that the names are different.

fspreiss, line 77: We could already make an entry in the respective CHANGELOG in some `[Unreleased]` section.

fspreiss (approved):
Thanks, @randombit!

frontend/ic_vetkeys/src/utils/utils.ts
randombit: While I'm sure I was the one who named the Rust function as well, I also thought the `_encoding` suffix was odd and omitted it here. It might be worthwhile (perhaps before we do the next release of the Rust + TS libraries) to go over both and try to harmonize naming wherever possible. (Opened CRP-2841 for this)

PR #143, opened 2025-06-11
chore: CRP-2806 prepare Motoko backend for publishing

fspreiss:
Adapts the Motoko backend folder so that it can be published on mops.one. In particular:

* Fixes the mops.toml
  * Corrects the syntax for the `keywords`
  * Adds a proper description
  * Removes the `repository` keyword because repositories located in sub-directories are not supported.
* Adds the repository location to the README
* Adds a change log

Also updates the date format in all changelogs to YYYY-MM-DD.


backend/mo/ic_vetkeys/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Change Log
+
+## [0.1.0] - 11-06-2025
```
randombit, line 3: ISO 8601 style (YYYY-MM-DD) is IMO preferable as it is not ambiguous, since nobody (per https://en.wikipedia.org/wiki/Calendar_date#Date_format) uses YYYY-DD-MM. In North American convention 11-06-2025 is Nov 6.

randombit (approved):
:+1:

PR #144, opened 2025-06-11
chore: CRP-2842 Add zeroization to VetKey and IbeSeed types

randombit:


backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -251,7 +251,7 @@ impl DerivedPublicKey {
 /// A VetKey is a valid BLS signature created for an input specified
 /// by the user
 ///
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, Zeroize, ZeroizeOnDrop)]
```
fspreiss, line 254: Note that in our [SecretArray](https://github.com/dfinity/ic/blob/3564b37939f037ba4d051ada88251c13954597d2/rs/crypto/secrets_containers/src/secret_array.rs#L1-L13) we go even further and box the array. Not sure if that is really needed here, but I wanted to point it out, so optionally, we could do the same here and for `IbeSeed`.

fspreiss (approved):
Thanks, @randombit!

PR #150, opened 2025-06-23
fix(basic_timelock_ibe): frontend bug in displaying bids

altkdf:
We reversed the lots based on their timestamp but used non-inverted indexing. This means that we displayed bid information as if the lots would not be reversed.

(This did not affect the correctness of the submitted bid AFAICT because that didn't use the wrong index variable.)

### DIFF
--- a/examples/basic_timelock_ibe/frontend/src/main.ts
+++ b/examples/basic_timelock_ibe/frontend/src/main.ts
@@ -319,7 +319,10 @@ async function listLots() {
             heading.textContent = "Open Lots";
             fragment.appendChild(heading);
 
-            openLots.lots.reverse().forEach((lot, index) => {
+            openLots.lots.reverse();
+            openLots.bidders.reverse();
+
+            openLots.lots.forEach((lot, index) => {
                 const lotDiv = document.createElement("div");
                 lotDiv.className = "lot";
                 const isCreator =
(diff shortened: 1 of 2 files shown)


examples/basic_timelock_ibe/frontend/src/main.ts
```diff
@@ -319,7 +319,10 @@ async function listLots() {
             heading.textContent = "Open Lots";
             fragment.appendChild(heading);
 
-            openLots.lots.reverse().forEach((lot, index) => {
```
fspreiss, line 322: We could further improve this by removing the implicit assumption that `OpenLotsResponse::lots` and `OpenLotsResponse::bidders` are in the same order. 

For example by changing the struct to `openlots: Vec<(LotInformation, Vec<Principal>)>`

fspreiss (approved):

examples/basic_timelock_ibe/frontend/src/main.ts
altkdf: Maybe. The downside would be that `Vec<Principal>` would not tell much about what it contains. We can add a type alias in Rust, but it would not be mapped to TS, so we would need a new type for `(LotInformation, Vec<Principal>)` or `Vec<Principal>`, which means some more complexity. So maybe the current state of affairs is not too bad. But I could do that if you have a strong preference.

fspreiss: No, it's fine to keep it as is!

PR #153, opened 2025-06-23
feat: CRP-2843 Change IBE ciphertext format so that streaming interfaces are possible

randombit:
This makes 3 changes to the IBE ciphertext format to make it possible/reasonable to support a streaming interface with large ciphertexts

* When we compute mask (`t`) we previously did so using `hashToScalar(header || seed || msg)`. Now we use `hashToScalar(header || seed || SHA-256(msg))`. The reason for this is because the hashToScalar interfaces we have available do not support init-update-finalize style interfaces common with regular hash functions. In the previous approach, if `msg` was say 4 GB, we would have to actually allocate a 4 GB + (len(header) + len(seed)) array, copy it over, and hash it in one go. Now we can maintain in the streaming state the values of the header and seed (both small and fixed length), and just compute the SHA-256 of msg as we go along.
* The order of the ciphertext serialization has changed from `(auth, maskedSeed, maskedMsg)` to `(maskedMsg, maskedSeed, auth)`. The order of the last two is irrelevant. But it is important that the masked message comes first because we can compute maskedMsg as we go along, without having to know the *entire* message value. While we do have to know the entire message to compute `maskedSeed` and `auth`. This allows for a streaming encryption interface. (Not added here)
* Increases the version of the IBE header

Certainly it's possible to handle both versions of the IBE scheme (that is after all the reason we have a version number) but given that the production key has not been released it seems better to just change it right away and not worry about compatibility.


frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -851,46 +852,56 @@ export class IbeSeed {
  */
 export class IbeCiphertext {
     readonly #header: Uint8Array;
-    readonly #c1: G2Point;
-    readonly #c2: Uint8Array;
-    readonly #c3: Uint8Array;
+    readonly #masked_msg: Uint8Array;
```
altkdf, line 855: off-topic: did we leave snake case here to be consistent with the Rust implementation?

altkdf (approved):
LGTM @randombit . Maybe @fspreiss wants to take a look as well.

We should also add this to the changelogs.

frontend/ic_vetkeys/src/utils/utils.ts
randombit: It looks like there is an eslint rule that checks this (https://eslint.org/docs/latest/rules/camelcase) we can just enable it and then not worry about the issue going forward.

randombit: I'll wait for #157 to land then rebase and fix the variable naming here

PR #156, opened 2025-06-23
feat: CRP-2826 CRP-2827 Provide a more consistent interface in the TS API regarding serialization and deserialization

randombit:


altkdf (approved):
Thanks @randombit , LGTM! Please also add to the changelogs.

frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -25,29 +25,24 @@ export class TransportSecretKey {
...
+    static random() {
+        return new TransportSecretKey(bls12_381.utils.randomPrivateKey());
+    }
+
+    /**
+     * Deserialize TransportSecretKey from a bytestring
      *
-     * For most applications, prefer using the static method random
+     * The passed value would typically be a string previously returned
+     * by calling getSecretKey on a randomly-created TransportSecretKey.
```
fspreiss, line 38: In Rust, we have `TransportSecretKey::serialize` while here in TS it is `TransportSecretKey.getSecretKey`. Not sure if it is intended to also unify this.

frontend/ic_vetkeys/src/utils/utils.ts
fspreiss, line 1: I noticed that for `IbeSeed` the serialization/deserialization methods are called `getByes` and `fromBytes`. Similar for Rust. Is this also something that should be adapted?

fspreiss (approved):
Thanks, @randombit!

fspreiss:
Do I understand correctly, that CRP-2826 was actually already addressed before the PR? That is, `VetKey` did have a `serialize` function already for both TS and Rust.

frontend/ic_vetkeys/src/utils/utils.ts
randombit: Probably makes sense to use serialize/deserialize always

randombit:
> Do I understand correctly, that CRP-2826 was actually already addressed before the PR? That is, VetKey did have a serialize function already for both TS and Rust.

No, TS was missing `VetKey.serialize`, this was added in this PR. There was the (equivalent) `getSignatureBytes` but this was thought maybe confusing for users who want the raw encoding but aren't using it for BLS signatures.

fspreiss:
> No, TS was missing VetKey.serialize, this was added in this PR.

Ah, indeed! I missed that! Thanks for clarifying!

PR #171, opened 2025-06-27
feat: CRP-2797 Add functions to access the production master key

randombit:
This allows performing key derivation completely offline, without having to interact with the IC.


frontend/ic_vetkeys/src/utils/utils.test.ts
```diff
@@ -91,6 +91,30 @@ test("MasterPublicKey derivation", () => {
...
 
+test("MasterPublicKey derivation using prod key", () => {
+    const masterKey = MasterPublicKey.productionKey();
+
+    // TODO: once the production key is enabled, replace this with test data generated
+    // using a mainnet canister
+
+    const canisterId = new TextEncoder().encode("rdmx6-jaaaa-aaaaa-aaadq-cai");
+
+    const canisterKey = masterKey.deriveKey(canisterId);
```
altkdf, line 102: In Rust, we have `deriive_canister_key` and `derive_subkey`, which looks nice. Are there any arguments for not having this in TS?

altkdf (commented):
LGTM @randombit ! Just one question. If we decide to have the same API in TS as in Rust, we could also do that in another PR.

frontend/ic_vetkeys/src/utils/utils.test.ts
randombit: No reason for this, I think that's just one of those things that evolved separately back when the Rust lib was in the monorepo

randombit:
Ah, nice. I realized `key_1` can already be used for deriving public keys, so I updated the tests for that to also use live data.

frontend/ic_vetkeys/src/utils/utils.test.ts
randombit: Renamed the functions to `deriveCanisterKey` and `deriveSubKey`

altkdf (approved):
Cool, thanks @randombit ! Please alsoo add to the changelog.

PR #186, opened 2025-07-03
test: CRP-2752 run frontend tests concurrently

altkdf:
Pass `sequence.concurrent` to `vitest`. This instructs `vitest` to run all tests concurrently and reduces the runtime of the current tests from ~60s to ~20s. The main reason of the longer runtime was that we made quite a few update calls to pocket IC, which are not super fast. This change is important to not get problems with CI when we add more tests.

I ran this 5 times locally and didn't observe any flakiness. If this will be the case in the future, e.g., if we add more tests, then we could either try to limit the maximum number of tests running concurrently or change the strategy how we parallelize tests.

### DIFF
--- a/frontend/ic_vetkeys/package.json
+++ b/frontend/ic_vetkeys/package.json
@@ -90,7 +90,7 @@
         "prettier": "prettier --write .",
         "prettier-check": "prettier --check .",
         "test_utils": "vitest utils",
-        "test": "npm run test:deploy_all && export $(cat .test1.env .test2.env | xargs) && vitest",
+        "test": "npm run test:deploy_all && export $(cat .test1.env .test2.env | xargs) && vitest --sequence.concurrent",
         "test:deploy_all": "npm run test:deploy_key_manager_canister && npm run test:deploy_encrypted_maps_canister",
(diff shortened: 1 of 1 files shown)


frontend/ic_vetkeys/package.json
```diff
@@ -90,7 +90,7 @@
         "prettier": "prettier --write .",
         "prettier-check": "prettier --check .",
         "test_utils": "vitest utils",
-        "test": "npm run test:deploy_all && export $(cat .test1.env .test2.env | xargs) && vitest",
+        "test": "npm run test:deploy_all && export $(cat .test1.env .test2.env | xargs) && vitest --sequence.concurrent",
```
fspreiss, line 93: Initially, for me the command `npm run test` failed because I don't have `vitest` installed (globally) on my machine. I believe we can improve this by using `npx` everywhre we call `vitest`. For example, like this `npx vitest --sequence.concurrent`.

See [this](https://vitest.dev/guide/#adding-vitest-to-your-project):
> By default, npx will first check if the command exists in the local project's binaries. If it is not found there, npx will look in the system's $PATH and execute it if found. If the command is not found in either location, npx will install it in a temporary location prior to execution.

fspreiss (approved):

frontend/ic_vetkeys/package.json
altkdf: I think `npm i && npm test` should work. AFAICT `package.json` has `vitest` as `devDependency` and `npm i` should install it (together with the other deps). If that doesn't work `npm i --include=dev` should fix it.

PR #194, opened 2025-07-07
feat: Use VetKDKeyId instead of MasterPublicKeyId

randombit:
Other interfaces including the management canister use VetKDKeyId to identify which VetKD master key to use. Use this type also for looking up hardcoded production keys instead of the one-off MasterPublicKeyId enum.

### DIFF
--- a/backend/rs/ic_vetkeys/src/utils/mod.rs
+++ b/backend/rs/ic_vetkeys/src/utils/mod.rs
@@ -154,15 +154,6 @@ pub enum PublicKeyDeserializationError {
     InvalidPublicKey,
 }
 
-/// Enumeration identifying the production master public key
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum MasterPublicKeyId {
-    /// The production key created in June 2025
-    Key1,
-    /// The test key created in May 2025
-    TestKey1,
-}
-
 #[derive(Clone, Debug, Eq, PartialEq)]
 /// A master VetKD public key
 pub struct MasterPublicKey {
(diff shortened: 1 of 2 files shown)


backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -218,14 +209,19 @@ impl MasterPublicKey {
...
     /// This allows performing public key derivation offline
-    pub fn production_key(key_id: MasterPublicKeyId) -> Self {
-        match key_id {
-            MasterPublicKeyId::Key1 => Self { point: *G2_KEY_1 },
-            MasterPublicKeyId::TestKey1 => Self {
-                point: *G2_TEST_KEY_1,
-            },
+    ///
+    /// Returns None if the provided key_id is not known
+    pub fn from_key_id(key_id: &VetKDKeyId) -> Option<Self> {
```
fspreiss, line 214: It is still unclear if we will support pocket-ic keys. In case we do, given that in pocket-ic the keys have the same names, `VetKDKeyId` alone isn't sufficient in telling if the user wants the mainnet key or the pocket-ic key. So then we either have to switch back to an enum with variants for mainnet keys (e.g., `Mainnet_Bls12_381_G2_Key1`) and for pocket-ic keys (e.g., `PocketIc_Bls12_381_G2_Key1`), or we use different methods for instantiating mainnet keys (e.g., `from_mainnet_key`) and pocket-ic keys (`from_pocket_ic_key`). The latter seems somewhat preferred, given that then we can use the `VetKDKeyId` struct as parameter.

Given this, we could name this method `from_mainnet_key` now, and then have the option to add pocket-ic support in the future without a breaking change.

fspreiss: 👍

PR #198, opened 2025-07-14
refactor: rename `derive_public_vetkey` to `derive_unencrypted_vetkey`

fspreiss:
Renames `derive_public_vetkey` to `derive_unencrypted_vetkey`.

Because this is a private function only used internally, the CHANGELOG is unchanged.

### DIFF
--- a/backend/rs/ic_vetkeys/src/utils/mod.rs
+++ b/backend/rs/ic_vetkeys/src/utils/mod.rs
@@ -863,8 +863,10 @@ pub mod management_canister {
 
     use super::*;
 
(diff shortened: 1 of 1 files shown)


randombit:
Clippys also fixed in #197

altkdf (approved):
Thanks, @fspreiss!

Just one remark. I don't fully understand this sentence. What does "otherwise" refer to?

```rs
// Therefore, this function shall only be used if the vetKey would otherwise be decrypted directly by the canister itself [...]
```

PR #200, opened 2025-07-21
chore: CRP-2888 add `rust-toolchain.toml` to improve reproducibility

altkdf:


rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "1.88.0"
```
randombit, line 2: a) Should we use some earlier version? In practice if we test only with 1.88, then it's not improbable that over time changes are introduced that preclude using it with earlier versions.

b) Regardless of version chosen, we should reflect this in [the cargo-advertised MSRV](https://doc.rust-lang.org/cargo/reference/rust-version.html#rust-version) as then attempting to use with an older version causes a nice "your rustc is too old for this crate" error rather than a potentially hard to understand compilation error.

altkdf: Does b) imply we should find the newest non-working compiler version?

randombit: Our current (implicit) MSRV is 1.85:

```
$ cargo +1.84 build
error: failed to download `hex-literal v1.0.0`

Caused by:
  unable to get packages from source

Caused by:
  failed to parse manifest at `/home/jack/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hex-literal-1.0.0/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.84.1 (66221abde 2024-11-19)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```

randombit: Using Rust 1.85 would also let us set Rust 2024 edition for all of the crates; I imagine that's why all of the RustCrypto crates are moving to this version as their MSRV

randombit: So `channel = "1.85.0"` plus changing `rust-version` in the various `Cargo.toml` to `1.85`.

randombit: lgtm

PR #202, opened 2025-07-22
chore: CRP-2848 Use noble-curves BLS signature verification function

randombit:


frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -396,21 +396,25 @@ export function verifyBlsSignature(
...
 ): boolean {
-    const negG2 = bls12_381.G2.ProjectivePoint.BASE.negate();
-    const oneGt = bls12_381.fields.Fp12.ONE;
+    const publicKeyBytes = pk.publicKeyBytes();
 
-    const signaturePt =
-        signature instanceof bls12_381.G1.ProjectivePoint
-            ? signature
-            : bls12_381.G1.ProjectivePoint.fromHex(signature);
+    const publicKeyAndMessage = new Uint8Array([...publicKeyBytes, ...message]);
```
altkdf, line 401: Maybe add a short comment why we pass here both the public key bytes and the message

altkdf (approved):

frontend/ic_vetkeys/src/utils/utils.ts
randombit: I added some references to the IETF draft explaining the augmented signature scheme

PR #208, opened 2025-07-28
ci: remove unnecessary `paths` and don't always install the heavy `candid-extractor`

altkdf:


altkdf:
In the examples we don't care about the local version of the libraries anymore because we use the public ones from `npmjs`/`crates.io`

PR #215, opened 2025-08-04
fix: enable the use of `KeyManager` and `EncryptedMaps` in `persistent actor`s

altkdf:
This PR extracts the state of the key manager/encrypted maps from the class object, since classes cannot be persisted using enhanced orthogonal persistence right now. Without this change, making canisters that use `KeyManager` or `EncryptedMaps` upgradable without losing data would be challenging.


backend/mo/canisters/ic_vetkeys_manager_canister/mops.toml
```diff
@@ -10,3 +10,4 @@ license = "Apache-2.0"
 
 [dependencies]
 base = "0.14.6"
+ic-vetkeys = "0.3.0"
```
altkdf, line 13: This is in line with the encrypted maps canister and avoids adjusting this canister due to changes in the backend library.

randombit (approved):

PR #219, opened 2025-08-11
perf: CRP-2900 Use optimized G2 generator point multiplication

randombit:
This saves 123M cycles during execution in a canister, as reported by the canbench results.

The implementation is just a port of the existing algorithm in https://github.com/dfinity/ic/blob/master/rs/crypto/internal/crypto_lib/bls12_381/type/src/lib.rs used for multiplication by fixed points in the replica.

### DIFF
--- a/backend/rs/ic_vetkeys/src/utils/mod.rs
+++ b/backend/rs/ic_vetkeys/src/utils/mod.rs
@@ -120,7 +120,11 @@ impl G2PrecomputedTable {
             s
         };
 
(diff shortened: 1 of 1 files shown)


randombit:
This adds two versions, one constant time and one variable time. The variable time is used only for public key derivation, IBE encryption and decryption both always use the constant time. We could use variable time for IBE in some contexts (eg auction finalization in a canister) but doing so improves the results at best by 10M cycles, and that's only when bumping the window size up to 6 bits, *and* it requires the user understand when it is or is not safe to use variable time execution. Given the savings are relatively small, it seemed to me better to use variable time only in cases where it's always safe.

PR #220, opened 2025-08-11
feat: CRP-2847 Add AES-GCM encryption helpers to the Rust library

randombit:
This matches the behavior of the functions in the TS frontend library

### DIFF
--- a/backend/rs/ic_vetkeys/src/utils/mod.rs
+++ b/backend/rs/ic_vetkeys/src/utils/mod.rs
@@ -373,6 +373,91 @@ impl VetKey {
     }
 }
 
+/// Key material derived from a VetKey
+#[derive(Clone, Zeroize, ZeroizeOnDrop)]
+pub struct DerivedKeyMaterial {
+    key: Vec<u8>,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+/// An error while encrypting
+pub enum EncryptionFailed {
+    /// The provided message was too long to be encrypted
+    PlaintextTooLong,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+/// An error while decrypting
+pub enum DecryptionFailed {
+    /// The ciphertext was too short to possibly be valid
+    MessageTooShort,
+    /// The GCM tag did not validate
+    InvalidCiphertext,
+}
+
+impl DerivedKeyMaterial {
+    /// Create a new DerivedKeyMaterial
+    pub fn new(vetkey: &VetKey) -> Self {
+        Self {
+            key: vetkey.serialize().to_vec(),
+        }
+    }
+
+    fn derive_aes_gcm_key(&self, domain_sep: &str) -> Vec<u8> {
+        derive_symmetric_key(&self.key, domain_sep, 32)
+    }
+
+    /// Encrypt a message
+    pub fn encrypt_message<R: rand::RngCore + rand::CryptoRng>(
+        &self,
+        message: &[u8],
+        domain_sep: &str,
+        rng: &mut R,
+    ) -> Result<Vec<u8>, EncryptionFailed> {
+        use aes_gcm::{aead::Aead, aead::AeadCore, Aes256Gcm, Key, KeyInit};
+        let key = self.derive_aes_gcm_key(domain_sep);
+        let key = Key::<Aes256Gcm>::from_slice(&key);
+        let nonce = Aes256Gcm::generate_nonce(rng);
+        let gcm = Aes256Gcm::new(key);
+
+        // The function returns an opaque `Error` with no details, but upon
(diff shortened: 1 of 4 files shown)


backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -373,6 +373,91 @@ impl VetKey {
...
+impl DerivedKeyMaterial {
+    /// Create a new DerivedKeyMaterial
+    pub fn new(vetkey: &VetKey) -> Self {
+        Self {
+            key: vetkey.serialize().to_vec(),
+        }
+    }
+
+    fn derive_aes_gcm_key(&self, domain_sep: &str) -> Vec<u8> {
+        derive_symmetric_key(&self.key, domain_sep, 32)
```
fspreiss, line 407: Do I understand correctly that the Rust implementation is compatible with the Typescript implementation regarding key-derivation/encryption/decryption, i.e., decrypting a ciphertext in the canister (with our Rust lib) that was encrypted in the browser (with our Typescript lib) would work? If so, if would make sense to explicitly mention this in the documentation.

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -373,6 +373,91 @@ impl VetKey {
...
+    pub fn encrypt_message<R: rand::RngCore + rand::CryptoRng>(
+        &self,
+        message: &[u8],
+        domain_sep: &str,
+        rng: &mut R,
+    ) -> Result<Vec<u8>, EncryptionFailed> {
+        use aes_gcm::{aead::Aead, aead::AeadCore, Aes256Gcm, Key, KeyInit};
+        let key = self.derive_aes_gcm_key(domain_sep);
+        let key = Key::<Aes256Gcm>::from_slice(&key);
+        let nonce = Aes256Gcm::generate_nonce(rng);
```
fspreiss, line 420: Maybe add a comment that this produces a (fixed-size) 12-byte nonce.

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -373,6 +373,91 @@ impl VetKey {
...
+        &self,
+        ctext: &[u8],
+        domain_sep: &str,
+    ) -> Result<Vec<u8>, DecryptionFailed> {
+        use aes_gcm::{aead::Aead, Aes256Gcm, Key, KeyInit};
+        let key = self.derive_aes_gcm_key(domain_sep);
+        let key = Key::<Aes256Gcm>::from_slice(&key);
+
+        // Minimum possible length 12 bytes nonce plus 16 bytes GCM tag
+        if ctext.len() < 12 + 16 {
```
fspreiss, line 447: nit: Following the security review, for the Typescript variant we [changed](https://github.com/dfinity/vetkeys/pull/146/files) the integer literals to named constants. I suggest we do the same here.

backend/rs/ic_vetkeys/src/utils/mod.rs
```diff
@@ -373,6 +373,91 @@ impl VetKey {
...
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+/// An error while decrypting
+pub enum DecryptionFailed {
+    /// The ciphertext was too short to possibly be valid
+    MessageTooShort,
+    /// The GCM tag did not validate
+    InvalidCiphertext,
+}
+
+impl DerivedKeyMaterial {
```
fspreiss, line 398: This is mainly a note to myself: One could argue that the intermediate `DerivedKeyMaterial` is only needed in the Typescript variant because there we need it because of the `CryptoKey` wrapper, and that the encryption/decryption functions could live directly on the `VetKey`.

However, having the intermediate `DerivedKeyMaterial` also here is Rust still seems to make sense because (1) then API-wise we are in line with Typescript and (2) it is more apparent/explicit that the encryption doesn't happen with the vetKey itself but with a key that is derived from it.

fspreiss (commented):
Thanks, @randombit!

backend/rs/ic_vetkeys/src/utils/mod.rs
randombit: Exactly so

I'll update the docs

backend/rs/ic_vetkeys/src/utils/mod.rs
randombit: Exactly so, this could have easily also been some freestanding functions, but having the functions have similar names/interfaces between the two libraries seemed like it would make a life a bit easier for developers

backend/rs/ic_vetkeys/src/utils/mod.rs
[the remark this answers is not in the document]
randombit: Opened CRP-2911

PR #234, opened 2025-09-10
fix(crypto): basic_ibe example: use persistent actor and Motoko core

fspreiss:
Adapts the basic_ibe example so that it works again with ICP Ninja, which recently upgraded to DFX 0.29.1, which ships with a new motoko compiler (0.16.1). The necessary changes for this were

* Declaring the actor as `persistent`
* Changing the `inboxes` map to the stable mo:core/Map

As the examples was touched anyway, the PR switches from the outdated `mo:base` to the new `mo:core` and fixes all compiler issues.

The Github workflow sets the `DFX_VERSION` environment variable, so that only this example uses the new DFX version 0.29.1. This also required to change the bind host from `localhost` to `127.0.0.1` as otherwise this new DFX version hits the following error:
```
Error: The replica returned an HTTP Error: Http Error: status 400 Bad Request, content type "text/plain; charset=utf-8", content: error: no_authority
details:
The request is missing the required authority information (e.g. 'Host' header).
```

The changes were tested in ICP Ninja with the following link:

[](http://icp.ninja/editor?g=https://github.com/dfinity/vetkeys/tree/franzstefan/CRP-2924-basic-ibe/examples/basic_ibe/motoko)

### DIFF
--- a/examples/basic_ibe/frontend/src/main.ts
+++ b/examples/basic_ibe/frontend/src/main.ts
@@ -226,32 +226,12 @@ async function displayMessages(inbox: Inbox) {
                     if ("Err" in result) {
                         alert("Error deleting message: " + result.Err);
                     } else {
-                        // Remove the message element from the DOM
-                        const messageElement = target.closest(".message");
-                        if (messageElement) {
-                            messageElement.remove();
-
-                            // If this was the last message, show the "no messages" message
-                            const messagesDiv =
-                                document.getElementById("messages")!;
-                            if (messagesDiv.children.length === 0) {
-                                const noMessagesDiv =
-                                    document.createElement("div");
-                                noMessagesDiv.className = "no-messages";
-                                noMessagesDiv.textContent =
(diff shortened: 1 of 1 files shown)


examples/basic_ibe/frontend/src/main.ts
```diff
@@ -226,33 +226,13 @@ async function displayMessages(inbox: Inbox) {
...
-                                const noMessagesDiv =
-                                    document.createElement("div");
-                                noMessagesDiv.className = "no-messages";
-                                noMessagesDiv.textContent =
-                                    "No messages in the inbox.";
-                                messagesDiv.appendChild(noMessagesDiv);
-                            }
-                        }
+                        // Re-load all messages to refresh message indices
+                        await showMessages();
```
fspreiss, line 230: This fixes a bug that was already there before this PR. We need to reload all messages because the frontend knows messages only by index, but the indices actually change when a message is deleted.

altkdf (approved):

PR #237, opened 2025-09-11
fix(crypto): password manager examples: use persistent actor

fspreiss:
Adapts the password_manager and password_manager_with_metadata examples so that they work again with ICP Ninja, which recently upgraded to DFX 0.29.1, which ships with a new motoko compiler (0.16.1). The necessary changes for this were

* Declaring the actor as `persistent` and fixing the compiler issues

The Github workflow sets the `DFX_VERSION` environment variable, so that only this example uses the new DFX version 0.29.1. This also required to change the bind host from `localhost` to `127.0.0.1` as otherwise this new DFX version hits the following error:
```
Error: The replica returned an HTTP Error: Http Error: status 400 Bad Request, content type "text/plain; charset=utf-8", content: error: no_authority
details:
The request is missing the required authority information (e.g. 'Host' header).
```

The changes were tested in ICP Ninja with the following link:

Password manager:
[](http://icp.ninja/editor?g=https://github.com/dfinity/vetkeys/tree/franzstefan/CRP-2924-password-managers/examples/password_manager/motoko)

Password manager with metadata:
[](http://icp.ninja/editor?g=https://github.com/dfinity/vetkeys/tree/franzstefan/CRP-2924-password-managers/examples/password_manager_with_metadata/motoko)

### DIFF
--- a/.github/workflows/examples-password-manager-with-metadata.yml
+++ b/.github/workflows/examples-password-manager-with-metadata.yml
@@ -13,6 +13,8 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
+env:
+  DFX_VERSION: 0.29.1
 jobs:
   examples-password-manager-with-metadata-rust-darwin:
     runs-on: macos-15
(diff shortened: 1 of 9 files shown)


randombit (approved):
lgtm

PR #366, opened 2026-04-14
chore(examples): migrate all 7 examples to icp-cli and @dfinity/vetkeys

marc0olo:
## Summary

This PR migrates all 7 example projects to use **ICP CLI** instead of dfx and **`@icp-sdk/core`** instead of `@dfinity/*`, and fixes a number of runtime issues discovered during testing.

> **Draft status**: The `@dfinity/vetkeys` package is currently referenced via a local `file:` path. Once `@dfinity/vetkeys` v0.5.0 is published to npm, this will be swapped to `^0.5.0` and this PR can be moved to ready-for-review.
>
> **icp.ninja badges**: All icp.ninja launch badges have been temporarily commented out. icp.ninja currently requires dfx; the badges will be re-enabled once icp.ninja supports icp-cli.

### Changes by commit

1. **`feat(examples): migrate all 7 example frontends to @icp-sdk/core`**
   - Replace `@dfinity/auth-client` + `@dfinity/principal` with `@icp-sdk/auth` + `@icp-sdk/core`
   - Add `dev:motoko` / `dev:rust` npm scripts, update vite configs, refresh `package-lock.json`
   - Update `gen_bindings.sh` scripts to use `icp-cli`
   - Update canister declaration files and `encrypted_maps_canister.ts` to match current backend interface
   - Remove deprecated `baseUrl` from `tsconfig.json`; rename `tailwind.config.cjs` → `.mjs` in password_manager examples (`"type":"module"`)

2. **`feat(examples): migrate all 7 example backends from dfx to icp-cli`**
   - Replace `dfx.json` with `icp.yaml` for all Motoko and Rust backends
   - Update `mops.toml` to pin `moc 1.5.0`, remove `--enhanced-orthogonal-persistence` flag (default since moc 0.15.0)
   - Update Rust backend for `password_manager_with_metadata` to current `ic-vetkeys` crate interface

3. **`fix(examples): update frontend source for @dfinity/vetkeys and @icp-sdk API changes`**
   - Rename `getBasicIbeCanister` → `getBasicIbeActor` (and equivalents) for naming consistency
   - Remove `window.global` polyfill and `as any` casts (no longer needed)
   - Fix actor caching in `encrypted_chat` auth store (create once on login, not per call)
   - Fix `DerivedKeyMaterial.fromCryptoKey()` missing `await` (method is now async)
   - Add required `associatedData` arg to `encryptMessage` / `decryptMessage`
   - Rewrite `encrypted_notes` `crypto.ts`: `deriveAesGcmCryptoKey` is now private; use `encryptMessage`/`decryptMessage` on `DerivedKeyMaterial` directly; cache `CryptoKey` in IndexedDB and reconstruct on retrieval (class instances are not structured-clone safe)
   - Replace `HttpAgent.createSync()` with async `HttpAgent.create()`; remove unused `hex_encode`

4. **`fix(ci): update CI workflows to use icp-cli instead of dfx`**
   - Replace `dfx start/deploy` with `icp network start / icp deploy`
   - Remove `DFX_VERSION` env var, add Node.js 22 setup step
   - Update provisioning scripts to install icp-cli

5. **`docs(examples): update documentation links and ICP CLI references`**
   - `internetcomputer.org/docs/` → `docs.internetcomputer.org/`
   - IC SDK install links → [ICP CLI](https://cli.internetcomputer.org)
   - Add `icp network stop` instructions to local deployment sections
   - Fix broken SPEC.md anchor in `encrypted_chat`

---

### `gen_bindings.sh` — `candid-extractor` guard

Each example's `gen_bindings.sh` script calls `make extract-candid` to refresh the static `.did` file from the compiled WASM, then feeds that file to `@icp-sdk/bindgen`. The Makefile target ends with a shell redirect: `candid-extractor ... > backend.did`.

In bash, the `>` redirect **opens and truncates the target file before the command runs**. If `candid-extractor` is not installed (it isn't — our provision scripts don't include it), the static `.did` file is wiped to zero bytes even though the command itself exits 127. `bindgen` then reads an empty file and generates a `did.js` with no exports, causing a vite build failure (`"idlFactory" is not exported`).

On `main` this was harmless: `gen_bindings.sh` used `dfx generate` which reads Candid from the deployed canister, not from the `.did` file on disk. With `bindgen`, the file is the explicit source of truth, so the truncation becomes a real failure.

The guard `if command -v candid-extractor >/dev/null 2>&1` skips the extraction entirely when the tool is absent, leaving the committed `.did` files intact for `bindgen` to read.

---

### Pre-existing `encrypted_chat` SPEC.md divergences (not introduced by this PR)

The `encrypted_chat` example is an unfinished prototype (README has a disclaimer). The following gaps between `SPEC.md` and the actual code predate this PR and exist on both `main` and this branch:

1. **`get_my_chats_and_time` → `get_my_chat_ids`**: SPEC defines an API returning `{chats, consensus_time_now}`, but backend/frontend use `get_my_chat_ids` returning only an array of tuples.
2. **Extra parameter in create chat APIs**: SPEC says `create_direct_chat(OtherParticipant, SymmetricKeyRotationMins)`, but both backend and frontend pass `messageExpirationDurationMinutes` as a third parameter during creation.
3. **`derive_chat_vetkey` VetKeyEpochId**: optional in code (`[] | [bigint]`), required in SPEC.
4. **Nonce type mismatch**: SPEC defines `Nonce = blob`, backend DID uses `bigint`.
5. **8-byte nonces**: SPEC specifies 16 bytes; frontend uses 8-byte nonces.

## Test plan

- [ ] `basic_ibe`: `icp network start -d && icp deploy` (Motoko + Rust), verify encrypt/decrypt flow
- [ ] `basic_bls_signing`: verify sign/verify flow
- [ ] `basic_timelock_ibe`: verify timelock decrypt at future epoch
- [ ] `encrypted_notes_dapp_vetkd`: add note, reload page (tests IndexedDB cache path), verify decrypt
- [ ] `encrypted_chat`: send message, verify decryption on both ends
- [ ] `password_manager`: add vault/password, verify encrypt/decrypt
- [ ] `password_manager_with_metadata`: same as above with metadata fields
- [ ] CI passes on all workflow jobs

🤖 Generated with [Claude Code](https://claude.com/claude-code)


fspreiss:
I assume this PR is now obsolete.

marc0olo:
@fspreiss yes, but I think we can use the individual example changes in future PRs on the examples repo once the new library is released.

at least we could point Claude to this PR when performing the actual migration. I tested all of the examples and it wasn't trivial to make everything work. (incl. the devserver setup for the frontend)

I would recommend to keep the branch alive until that is done 😅

PR #377, opened 2026-04-21
chore: remove examples

andreacerulli:
The examples from the vetkeys repo are being moved to dfinity/examples. This makes it easier to ensure they can be built using icp-ninja.


.gitignore
```diff
@@ -93,3 +93,4 @@ docs/
 *.mops
 
 **/.canbench
+frontend/ic_vetkeys/.test*.env
```
fspreiss, line 96: not sure what that is, but looks reasonable

.github/workflows/examples-basic-bls-signing.yml
fspreiss, line 1: Optionally we could move the workflows temporarily to the examples repo until they are integrated into the [.github/workflows/ninja_pr_checks.yml](https://github.com/dfinity/examples/blob/master/.github/workflows/ninja_pr_checks.yml), but probably it is better to use the time to get the ninja_pr_checks working.

fspreiss (approved):

.github/workflows/examples-basic-bls-signing.yml
andreacerulli: I spoke with Marco and his preference is to get rid of dfx everywhere, however this would break icp-ninja IIUC. So we have to check whether having them running on icp-ninja is actually a priority. In that case, we could move these workflows there. This is however independent on this pr.

PR #378, opened 2026-04-21
feat: migrate @dfinity/* to @icp-sdk/core and icp-cli

andreacerulli:
Migrate the frontend library and backend infrastructure from deprecated @dfinity/* packages and dfx to @icp-sdk/core and icp-cli.

Frontend library (frontend/ic_vetkeys):
- Replace @dfinity/{agent,principal,candid,identity} with @icp-sdk/core
- Regenerate declarations with @icp-sdk/bindgen@0.3.0
- Replace createActor wrappers with direct Actor.createActor calls
- Update test host ports and add canister ID null guards
- Reformat tsconfig, vite.config, eslint config

Backend infrastructure:
- Replace dfx.json with icp.yaml for all backend canisters
- Update Makefiles for icp-cli compatible builds
- Update test key from dfx_test_key to test_key_1
- Update CI workflows for icp-cli

CI:
- Update frontend.yml and backend-motoko.yml for icp-cli
- Update provision scripts
- Add .icp to .gitignore

PR #381, opened 2026-04-22
feat: migrate @dfinity/* to @icp-sdk/core and icp-cli (#378)

andreacerulli:
Migrate the frontend library and backend infrastructure from deprecated @dfinity/* packages and dfx to @icp-sdk/core and icp-cli.

Frontend library (frontend/ic_vetkeys):
- Replace @dfinity/{agent,principal,candid,identity} with @icp-sdk/core
- Regenerate declarations with @icp-sdk/bindgen@0.3.0
- Replace createActor wrappers with direct Actor.createActor calls
- Update test host ports and add canister ID null guards
- Reformat tsconfig, vite.config, eslint config

Backend infrastructure:
- Replace dfx.json with icp.yaml for all backend canisters
- Update Makefiles for icp-cli compatible builds
- Update test key from dfx_test_key to test_key_1
- Update CI workflows for icp-cli

CI:
- Update frontend.yml and backend-motoko.yml for icp-cli
- Update provision scripts
- Add .icp to .gitignore


marc0olo (approved):
LGTM from my side. I compared this against my earlier PR #365 (which I had fully tested locally), and the changes to all overlapping files are **identical** except for two intentional differences:

- `.github/workflows/backend-motoko.yml`: #381 adds a `setup-node@v4` step for both Linux and Darwin jobs, which #365 was missing — this is a good addition.
- `.gitignore`: #365 also added `frontend/ic_vetkeys/.test*.env`, but that entry turns out to be unnecessary (no such files are generated), so omitting it here is fine.

Approving. ✅

randombit (approved):

PR #384, opened 2026-04-22
chore: migrate npm package from @dfinity/vetkeys to @icp-sdk/vetkeys

### LINKED ISSUE #382: chore: migrate @dfinity/vetkeys to @icp-sdk/vetkeys
## Summary

Rename the published npm package from `@dfinity/vetkeys` to `@icp-sdk/vetkeys` and migrate the release workflow to follow the [icp-js-auth](https://github.com/dfinity/icp-js-auth) pattern.

## Tasks

- [ ] Rename `"name"` in `frontend/ic_vetkeys/package.json` to `@icp-sdk/vetkeys`
- [ ] Update `homepage` and any README/docs prose referencing the old package name
- [ ] Switch from npm to pnpm (required to align with `dfinity/ci-tools` actions)
- [ ] Add `pnpm-lock.yaml` and remove `package-lock.json`
- [ ] Add Commitizen config (`.cz.toml`)
- [ ] Add migration note to `CHANGELOG.md` for the 0.5.0 entry explaining that versions 0.1.0–0.4.0 were published as `@dfinity/vetkeys`
- [ ] Replace `.github/workflows/publish-frontend.yml` with `create-release-pr.yml` + `release.yml` following the icp-js-auth pattern (simplified tag format `1.2.3`, pnpm, npm provenance, auto-generated GitHub release notes)
- [ ] Confirm `NPM_TOKEN` in the `release` environment has publish rights to the `@icp-sdk` npm org (prerequisite — may need coordination)
- [ ] After first successful publish: deprecate `@dfinity/vetkeys` on npm via `npm deprecate @dfinity/vetkeys@"*" "This package has moved to @icp-sdk/vetkeys"` (requires `@dfinity` npm org access)

marc0olo:
## Summary

- Renames npm package from \`@dfinity/vetkeys\` to \`@icp-sdk/vetkeys\` and switches from npm to pnpm
- Replaces \`publish-frontend.yml\` with \`release-npm.yml\`: tag-triggered (\`npm/X.Y.Z\`), with version consistency check, dependency audit, provenance attestation, and \`workflow_dispatch\` dry-run mode — all pinned to commit SHAs
- Renames \`publish-ic-vetkeys.yml\` to \`release-rust.yml\` (no functional changes)
- Adds \`RELEASING.md\` with documented release process for all three libraries (npm fully documented; Rust and Motoko marked TODO)
- Adds \`@noble/curves\` and \`@noble/hashes\` as explicit \`devDependencies\` — pnpm does not hoist transitive deps, so they must be declared directly
- Drops \`isomorphic-fetch\` and \`happy-dom\` (Node.js 22 provides native Web Crypto and \`fetch\`); changes vitest default environment from \`happy-dom\` to \`node\`
- Bumps \`typescript-eslint\` to \`^8.59.0\` for TypeScript 5.9 peer-dep support
- Tightens internal \`Uint8Array\` generics to \`Uint8Array<ArrayBuffer>\` for TypeScript 5.9 compatibility; replaces unsafe casts with defensive copies (\`asBytes\`, \`getDerivedKeyMaterialOrFetchIfNeeded\`, \`VetKey\` constructor)
- Updates \`CHANGELOG.md\` with migration note and TypeScript 5.9 fix entry

> **Note:** deprecated \`@noble/curves\`/\`@noble/hashes\` API fixes and JSDoc corrections are split into a separate PR (#386, already merged) to keep concerns isolated.

## Release flow (once merged into main)

Follow the steps in \`RELEASING.md\` to cut the 0.5.0 release.

## Notes

- Deprecating \`@dfinity/vetkeys\` on npm (requires \`@dfinity\` org access) is a separate post-release step
- The README module doc links (\`dfinity.github.io/vetkeys/modules/…\`) are verified via local typedoc build and will resolve once the docs are deployed as part of the release (step 7 in \`RELEASING.md\`)

Closes #382


package.json
```diff
@@ -1,12 +1,6 @@
...
     "private": true,
-    "workspaces": [
-        "frontend/ic_vetkeys"
-    ],
     "scripts": {
-        "build": "npm run build --workspaces --if-present; npm run bundle --workspaces --if-present"
-    },
-    "devDependencies": {
-        "svelte-hmr": "^0.16.0"
+        "build": "pnpm -r run build"
```
fspreiss: For me as a non-expert on this, `--recursive` instead of `-r` would be more clear, but both are fine for me.

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,93 @@
...
+    runs-on: ubuntu-latest
+    environment:
+      name: release
+      url: https://www.npmjs.com/package/@icp-sdk/vetkeys
+    permissions:
+      contents: write
+      id-token: write
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```
fspreiss: could use the latest v6.0.2 here and in publish-frontend-docs.yml

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,93 @@
...
+
+      - name: Setup PNPM
+        uses: dfinity/ci-tools/actions/setup-pnpm@main
```
fspreiss: What is the advantage of using this custom pnpm setup action? It is also inconsistent with other workflows (frontend.yml and publish-frontend.yml), that use the pnpm/action-setup, and the version is not pinned.
Also, will this really work? According to the action README it needs a to know a node version. The default file (.node-version) doesn't exist and also no node_version is explicitly passed, so I'm assuming this will fail at runtime.

Update: OK, it seems the advantage is that it sets up _both_ pnpm _and_ node and also configures the caching correctly. I'm fine with that, but we need to provide the node version somehow to make it work. Also, maybe we should use this then also in frontend.yml and publish-frontend.yml.

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Create Github release
+        if: ${{ github.event.inputs.dry-run != 'true' }}
+        uses: ncipollo/release-action@440c8c1cb0ed28b9f43e4d1d670870f059653174 # v1.16.0
+        with:
+          artifacts: 'frontend/ic_vetkeys/icp-sdk-vetkeys-${{ steps.ver.outputs.version }}.tgz'
+          bodyFile: 'RELEASE_NOTES.md'
+          tag: '${{ github.ref_name }}'
+          commit: ${{ github.sha }}
+          prerelease: ${{ steps.ver.outputs.is_beta == 'true' }}
+          makeLatest: ${{ steps.ver.outputs.is_beta == 'false' }}
```
fspreiss: I'm wondering what the added value of a Github Release here is. We haven't use this so far. Won't people just go look at npmjs.org to see the published versions? If we use Github Releases, we also have to make sure things keep in sync, which adds overhead/complexity.

Also, given that "this repo ships three independent libraries" (quote from RELEASING.md), having Github Releases only for the npm library is somewhat inconsistent because for the Rust library there is no Github Release created (see release-rust.yml).

My impression is, we can do without Github releases.

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Publish to npm
+        working-directory: frontend/ic_vetkeys
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NPM_CONFIG_PROVENANCE: 'true'
+        run: |
+          release_cmd="pnpm publish --verbose --access public --no-git-checks"
+          if [ "${{ steps.ver.outputs.is_beta }}" = 'true' ]; then
+            release_cmd="$release_cmd --tag beta"
+          fi
```
fspreiss: What is the motivation for introducing _beta_ releases? We didn't have a need for this so far. Seems like it just adds unnecessary complexity.

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,93 @@
...
+      contents: write
+      id-token: write
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Setup PNPM
+        uses: dfinity/ci-tools/actions/setup-pnpm@main
+
+      - name: Determine version
```
fspreiss: The old publish-frontend.yml also had a version consistency check (`Verify package.json version` step) that ensured that the package version and the release version are the same. Seems worth to keep this, no?

frontend/ic_vetkeys/src/utils/utils.ts
```diff
@@ -354,18 +354,21 @@ export function hashToScalar(input: Uint8Array, domainSep: string): bigint {
 /**
  * @internal helper for data encoding
  */
-function asBytes(input: Uint8Array | string): Uint8Array {
+function asBytes(input: Uint8Array | string): Uint8Array<ArrayBuffer> {
     if (typeof input === "string") {
         return new TextEncoder().encode(input);
     } else {
-        return input;
+        return input as Uint8Array<ArrayBuffer>;
```
fspreiss: Is this really sound? Here we don't know that `input` is really a `Uint8Array<ArrayBuffer>`.

Maybe something like
```
function asBytes(input: Uint8Array | string): Uint8Array<ArrayBuffer> {
    if (typeof input === "string") {
        return new TextEncoder().encode(input);
    }
    const out = new Uint8Array(input.length);
    out.set(input);
    return out;
}
```
would make it sound.

frontend/ic_vetkeys/src/encrypted_maps/index.ts
```diff
@@ -614,7 +614,7 @@ export class EncryptedMaps {
     ): Promise<DerivedKeyMaterial> {
         const cachedRawDerivedKeyMaterial: CryptoKey | undefined = await get([
             mapOwner.toString(),
-            mapName,
+            mapName as Uint8Array<ArrayBuffer>,
```
fspreiss: Also here I'm not sure this is sound, at least not when compiling with TS 5.9.

Maybe something like this
```
const safeMapName = new Uint8Array(mapName);  // copies into fresh ArrayBuffer
const cached = await get([mapOwner.toString(), safeMapName]);
...
await set([mapOwner.toString(), safeMapName], derivedKeyMaterial.getCryptoKey());
```

frontend/ic_vetkeys/README.md
```diff
@@ -8,23 +8,23 @@ This package contains three entry points:
 
 <br>
 
-[`ic_vetkeys`](https://dfinity.github.io/vetkeys/modules/_dfinity_vetkeys.html)
+[`@icp-sdk/vetkeys`](https://dfinity.github.io/vetkeys/)
```
fspreiss: I'd leave the link target what it was, so that the user jumps directly to the relevant section in the docs.

Same for the other two links in this file.

.github/workflows/release-npm.yml
marc0olo: I agree, there is not much added value and we would introduce inconsistency here. tags were already in place and could benefit from some documentation I guess so that somebody who is actually interested could lookup the code of a specific release.

I will remove that logic.

.github/workflows/release-npm.yml
marc0olo: mainly just replicating the logic from the other JS repos. I agree, we don't need it. I will remove it.

marc0olo:
**RE: README doc links** (addressing @fspreiss's comment)

The three module links in `README.md` now point to specific typedoc sections:
- `modules/_icp-sdk_vetkeys.html`
- `modules/_icp-sdk_vetkeys_key_manager.html`
- `modules/_icp-sdk_vetkeys_encrypted_maps.html`

These URLs were verified via a local `typedoc` build from the current branch — the generated filenames match exactly. They are currently returning 404 on the live site because the API docs haven't been redeployed since the package was renamed from `@dfinity/vetkeys` to `@icp-sdk/vetkeys` in PR #386. The docs deployment workflow (`publish-frontend-docs.yml`) is manual-only, so they'll go live once the workflow is triggered with the `npm/0.5.0` tag as part of the release.

I've added this as step 7 in `RELEASING.md` so it isn't missed.

frontend/ic_vetkeys/src/utils/utils.ts
fspreiss: This cast actually also seems unsound. This is not from your changes, but the following would be a defensive copy:

```
this.#bytes = new Uint8Array(pt.toBytes(true));
```

Would be great if you could also change this as part of this PR.

fspreiss (approved):
Thanks, @marc0olo!

PR #393, opened 2026-05-12
fix: flaky should_get_owned_map_names test

mraszyk:
---
  fix: correct should_get_owned_map_names test to handle duplicate map names
  
  random_map_name generates ByteBufs of length 0–31 bytes. With a specific RNG seed, the same map name can be generated in two different loop iterations. The test's Vec-based
  expected_map_names counted the duplicate, but get_owned_non_empty_map_names internally deduplicates via a HashSet, causing the assertion left == right to fail with left: 5, right: 6.

This PR fixes the issue by using BTreeSet (HashSet cannot be used because ByteBuf does not derive Hash) instead of Vec in the test, too.


pierugo-dfinity (commented):
I don't have much context unfortunately, but if the canister deduplicates map names, couldn't we also deduplicate them in this test, turning `expected_map_names` into a `HashSet`?

mraszyk:
> I don't have much context unfortunately, but if the canister deduplicates map names, couldn't we also deduplicate them in this test, turning `expected_map_names` into a `HashSet`?

Good idea! Done.

pierugo-dfinity (approved):

Issue #423, opened 2026-07-29
EncryptedMaps canister generators: allow excluding endpoints for wrap-and-extend use cases (e.g. password_manager_with_metadata)


marc0olo:
## Summary

The EncryptedMaps canister generators — the Rust `ic_vetkeys::export_encrypted_maps_canister!` macro (added in #404, shipped in `ic-vetkeys` 0.8.0) and the Motoko `EncryptedMapsCanister` mixin (#405) — are **all-or-nothing**: they emit the complete standard endpoint set with no supported way to omit or replace individual endpoints. This makes them a poor fit for the common "reuse the standard interface but wrap a few endpoints and keep some side state" pattern, and — more importantly — forces adopters of that pattern to *not use the generator at all*, or to expose endpoints that break their own invariants.

This issue proposes an opt-in way to generate the standard set **minus a chosen subset**, and asks whether the Motoko mixin can already achieve this.

## Motivating example

`password_manager_with_metadata` in [dfinity/examples](https://github.com/dfinity/examples) (both languages):

- [rust/vetkeys/password_manager_with_metadata/backend/src/lib.rs](https://github.com/dfinity/examples/blob/master/rust/vetkeys/password_manager_with_metadata/backend/src/lib.rs)
- [motoko/vetkeys/password_manager_with_metadata/backend/app.mo](https://github.com/dfinity/examples/blob/master/motoko/vetkeys/password_manager_with_metadata/backend/app.mo)

It stores, alongside each encrypted value, a second **backend-authoritative** metadata record (`creation_date`, `last_modification_date`, `number_of_modifications`, `last_modified_principal`, `tags`, `url`). These fields are computed and maintained by the canister (clock, caller principal, a counter), so they can't simply be folded into the client-encrypted blob — they need their own map. That gives the canister two stores that must stay 1:1:

```
EncryptedMaps:  (owner, name, key) -> ciphertext
metadata:       (owner, name, key) -> PasswordMetadata
```

The canister writes both in one call (`insert_encrypted_value_with_metadata` / `remove_encrypted_value_with_metadata`), and the read path (`get_encrypted_values_for_map_with_metadata`) zips the two together.

## The problem

[...]

PR #424, opened 2026-07-29
feat(rust): custom_value_endpoints form for export_encrypted_maps_canister!

marc0olo:
Implements the Rust half of #423.

## What changes
`export_encrypted_maps_canister!` gains a second form:

```rust
// Full canister (unchanged — the 95% case):
ic_vetkeys::export_encrypted_maps_canister!("app", [memory(0), memory(1), memory(2), memory(3)]);

// New: control-plane + state + lifecycle + accessors, but NO value endpoints:
ic_vetkeys::export_encrypted_maps_canister!("app", [memory(0), memory(1), memory(2), memory(3)], custom_value_endpoints);
```

The `custom_value_endpoints` form generates the 8 control-plane endpoints, the stable state, `#[init]`/`#[post_upgrade]`, and two accessors — but **omits the 7 value read/write endpoints**. Both forms now emit:

```rust
fn with_encrypted_maps<R>(f: impl FnOnce(&EncryptedMaps<AccessRights>) -> R) -> R;
fn with_encrypted_maps_mut<R>(f: impl FnOnce(&mut EncryptedMaps<AccessRights>) -> R) -> R;
```

## Why
The generators are all-or-nothing today, so the common "wrap-and-extend" pattern — a canister that stores something linked to each value (e.g. a metadata row per entry) and must own the value endpoints to keep the two stores consistent — can't use the macro at all and hand-writes ~200 lines. `custom_value_endpoints` lets it generate the safe control-plane passthroughs and write only its custom value endpoints, reusing the library's vetKD/crypto/ACL via the accessors.

**Coarse (whole value data-plane) instead of a fine-grained `exclude: [list]`, deliberately** — it's one unmissable decision rather than a list you can under-specify. (The `exclude` list in #423's own body proved the trap: it listed 6 names and omitted `get_all_accessible_encrypted_maps`, which returns raw values via `EncryptedMapData.keyvals` — a leak.) It's also forward-safe: a value endpoint added to the library later is hidden from adopters automatically, whereas a baked-in exclude list would silently re-expose it. Full rationale and the alternatives considered are in the [#423 thread](https://github.com/dfinity/vetkeys/issues/423).

## How it simplifies adoption (`password_manager_with_metadata`)
Today that example (~250 lines) forgoes the generator and hand-writes 11 endpoints; ~120–130 of those lines are the 8 passthroughs + init/post_upgrade/state setup that the generator already produces. With `custom_value_endpoints` those collapse to a 3-line macro call and only the 3 genuine `*_with_metadata` endpoints stay hand-written — roughly halving the file, keeping the passthroughs frontend-compatible by construction, and never emitting the raw value mutators (so the value↔metadata invariant can't be bypassed at the boundary).

## Which endpoints, and the rationale
- **Omitted (value data-plane, 7):** writes `insert_encrypted_value`, `remove_encrypted_value`, `remove_map_values`; reads `get_encrypted_value`, `get_encrypted_values_for_map`, `get_all_accessible_encrypted_values`, `get_all_accessible_encrypted_maps`.
- Only the **3 writes** can break an adopter invariant (correctness). The **4 reads** are dropped for API-surface hygiene (no metadata-less value view beside the wrapped one). Documented in the macro rustdoc.
- **Kept (control-plane, 8):** none read or write values — `set_user_rights`/`remove_user` touch only the ACL, no value cascade — so they're always safe to emit.

## Interface / release impact
- The **full** form is unchanged. Endpoints are factored into inner macros, but `export_candid!` emits methods **alphabetically**, so the generated `.did` is identical — verified by diffing the extracted candid (method set identical) and by interface-sync regenerating declarations from the committed `.did` (untouched). **No frontend files or `.did` change; no frontend release.**
- Additive/minor change. Recorded under an **Unreleased** changelog heading; the version number is assigned at release time (no `Cargo.toml` bump in this PR).

## Docs
- **Macro rustdoc** documents both forms, the omitted set with the write=correctness / read=hygiene rationale, the accessors, a metadata example, and the escape hatch if a write-only-wrap-keeping-reads need appears.
- **Crate README** (the docs.rs landing page) gains a "Ready-made EncryptedMaps canister" section pointing to the macro and both forms — it previously didn't mention the generator at all.
- **Reference-canister READMEs**: the new custom canister gets its own README, cross-linked from the full reference canister's README.

## Implementation & verification
- Factored the macro into `__export_encrypted_maps_{common,control_plane_endpoints,value_endpoints}!` helpers composed by the two public arms.
- Added a reference canister `ic-vetkeys-encrypted-maps-custom-canister` that dogfoods the new form + accessors; wired into the CI wasm build.
- Verified: full-form `.did` identical; encrypted_maps reference tests 19/19; doctests pass and `cargo doc -D warnings` clean; `cargo fmt --check` and CI-style `cargo clippy -- -Dwarnings` clean; custom canister builds host + wasm.

Motoko gets the mirrored layered mixin in a separate PR.

### DIFF
--- a/backend/rs/canisters/ic_vetkeys_encrypted_maps_custom_canister/src/lib.rs
+++ b/backend/rs/canisters/ic_vetkeys_encrypted_maps_custom_canister/src/lib.rs
@@ -41,6 +41,15 @@ fn to_blob(bytes: ByteBuf) -> Result<Blob<32>, String> {
 // A canister-owned value write, wrapping the library's own via the accessor.
 // This is where linked side-state (metadata, counters, …) would be maintained
 // in the same call — kept minimal here.
+//
+// Parameters (matching the EncryptedMaps model):
+// - `map_owner`: the principal that owns the map; together with `map_name` it
+//   identifies the map. Maps are namespaced per owner.
+// - `map_name`: the map's name within the owner's namespace (max 32 bytes).
+// - `map_key`: the key of the entry within that map (max 32 bytes).
+// - `value`: the already client-side-encrypted value to store.
+// Returns the previous value at that key, if any. Access is checked against the
+// caller (`msg_caller`) by the library.
 #[ic_cdk::update]
 fn insert_encrypted_value_custom(
     map_owner: Principal,
@@ -59,6 +68,10 @@ fn insert_encrypted_value_custom(
 }
 
 // A canister-owned value read via the shared-reference accessor.
+//
+// `map_owner` + `map_name` identify the map, `map_key` the entry within it (see
+// `insert_encrypted_value_custom`). Returns the encrypted value at that key, if
+// any; access is checked against the caller by the library.
 #[ic_cdk::query]
 fn get_encrypted_value_custom(
     map_owner: Principal,
--- a/backend/rs/canisters/ic_vetkeys_encrypted_maps_custom_canister/src/lib.rs
+++ b/backend/rs/canisters/ic_vetkeys_encrypted_maps_custom_canister/src/lib.rs
@@ -1,0 +1,77 @@
+// Reference canister for the `custom_value_endpoints` form of
+// `export_encrypted_maps_canister!`. It generates the control-plane endpoints
+// plus the state, lifecycle hooks, and the `with_encrypted_maps`/`_mut`
+// accessors, but NOT the value read/write endpoints — the canister exposes its
+// own. This is the pattern an app uses when it keeps state linked to each value
+// (e.g. a metadata row per entry) and must own the value endpoints to keep that
+// state consistent.
+//
+// This minimal example just re-implements a plain insert/get against the
+// accessor to exercise the surface; a full linked-metadata example lives in
+// `password_manager_with_metadata` in dfinity/examples.
+use candid::Principal;
+use ic_stable_structures::memory_manager::{MemoryId, MemoryManager, VirtualMemory};
+use ic_stable_structures::storable::Blob;
+use ic_stable_structures::DefaultMemoryImpl;
+use ic_vetkeys::types::{ByteBuf, EncryptedMapValue};
+use std::cell::RefCell;
+
+type Memory = VirtualMemory<DefaultMemoryImpl>;
+
+thread_local! {
+    static MEMORY_MANAGER: RefCell<MemoryManager<DefaultMemoryImpl>> =
+        RefCell::new(MemoryManager::init(DefaultMemoryImpl::default()));
+}
+
+fn memory(id: u8) -> Memory {
+    MEMORY_MANAGER.with(|m| m.borrow().get(MemoryId::new(id)))
+}
+
+// Control plane + state + lifecycle + accessors; no value endpoints.
+ic_vetkeys::export_encrypted_maps_canister!(
+    "encrypted_maps_custom_dapp",
+    [memory(0), memory(1), memory(2), memory(3)],
+    custom_value_endpoints,
+);
+
+fn to_blob(bytes: ByteBuf) -> Result<Blob<32>, String> {
+    Blob::try_from(bytes.as_ref()).map_err(|_| "too large input".to_string())
+}
+
+// A canister-owned value write, wrapping the library's own via the accessor.
+// This is where linked side-state (metadata, counters, …) would be maintained
+// in the same call — kept minimal here.
+#[ic_cdk::update]
+fn insert_encrypted_value_custom(
+    map_owner: Principal,
+    map_name: ByteBuf,
+    map_key: ByteBuf,
+    value: EncryptedMapValue,
+) -> Result<Option<EncryptedMapValue>, String> {
+    with_encrypted_maps_mut(|encrypted_maps| {
+        encrypted_maps.insert_encrypted_value(
+            ic_cdk::api::msg_caller(),
+            (map_owner, to_blob(map_name)?),
+            to_blob(map_key)?,
+            value,
+        )
+    })
+}
+
+// A canister-owned value read via the shared-reference accessor.
+#[ic_cdk::query]
+fn get_encrypted_value_custom(
+    map_owner: Principal,
+    map_name: ByteBuf,
+    map_key: ByteBuf,
+) -> Result<Option<EncryptedMapValue>, String> {
+    with_encrypted_maps(|encrypted_maps| {
+        encrypted_maps.get_encrypted_value(
+            ic_cdk::api::msg_caller(),
+            (map_owner, to_blob(map_name)?),
+            to_blob(map_key)?,
+        )
+    })
+}
+
+ic_cdk::export_candid!();
--- a/backend/rs/ic_vetkeys/src/encrypted_maps/canister.rs
+++ b/backend/rs/ic_vetkeys/src/encrypted_maps/canister.rs
@@ -1,17 +1,31 @@
-//! Macro that generates a complete EncryptedMaps canister.
+//! Macro that generates an EncryptedMaps canister.
 //!
 //! [`export_encrypted_maps_canister!`] expands — in the calling crate — to the
-//! `#[init]`, the stable state, and every `#[query]`/`#[update]` endpoint, so an
-//! adopter's `lib.rs` can be a few lines instead of ~200 lines of hand-written
-//! boilerplate. Because the macro is the single source of the endpoint set, the
-//! exposed Candid interface is exactly the one the `@icp-sdk/vetkeys` frontend
-//! expects, by construction.
+//! `#[init]`/`#[post_upgrade]`, the stable state, and the `#[query]`/`#[update]`
+//! endpoints, so an adopter's `lib.rs` can be a few lines instead of ~200 lines
+//! of hand-written boilerplate. Because the macro is the single source of the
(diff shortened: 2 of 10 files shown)


backend/rs/ic_vetkeys/src/encrypted_maps/canister.rs
```diff
@@ -23,15 +37,15 @@
 ///
 /// The `#[init]` takes the vetKD key name (e.g. `"test_key_1"` locally,
 /// `"key_1"` on mainnet). The macro injects the canister entry points into the
-/// invoking module — the `#[init]`/`#[post_upgrade]` and every `#[query]`/
-/// `#[update]` endpoint (plus a private thread-local and helpers) — but it does
+/// invoking module — the `#[init]`/`#[post_upgrade]` and the `#[query]`/
+/// `#[update]` endpoints (plus a private thread-local and helpers) — but it does
 /// **not** bind any common names in your scope (all library types are imported
 /// under unique aliases), so it won't clash with your own `use` imports. It
```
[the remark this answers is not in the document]
marc0olo: Good catch — applied. Updated the rustdoc: the 'does not bind common names' claim was really about the aliased *type* imports; it now states plainly that the macro also defines the endpoint functions and the `with_encrypted_maps`/`with_encrypted_maps_mut` accessors as items in the invoking module, so adopters shouldn't define their own items with those names.

backend/rs/ic_vetkeys/src/encrypted_maps/canister.rs
```diff
@@ -127,6 +283,38 @@ macro_rules! export_encrypted_maps_canister {
...
+        ///
+        /// For hand-written endpoints that reuse the library's vetKD / crypto /
+        /// access-control logic. Traps if called before `#[init]` (it never is —
+        /// the state is set up in `#[init]`/`#[post_upgrade]`).
+        #[allow(dead_code)]
+        fn with_encrypted_maps<__EmR>(
+            f: impl FnOnce(&__EmEncryptedMaps<__EmAccessRights>) -> __EmR,
+        ) -> __EmR {
+            ENCRYPTED_MAPS.with_borrow(|encrypted_maps| f(encrypted_maps.as_ref().unwrap()))
+        }
```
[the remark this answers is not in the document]
marc0olo: Applied both suggestions to `with_encrypted_maps` (and `with_encrypted_maps_mut`): made them `pub(crate)` so they're reachable from submodules of the invoking crate, and replaced `unwrap()` with `expect("EncryptedMaps is not initialized (called before #[init]/#[post_upgrade])")` for a clear trap. (Private items are already visible to descendant modules, but `pub(crate)` removes that footgun regardless of where the macro is invoked.) The generated endpoints keep their existing `unwrap()`s to keep the full-form diff minimal.

backend/rs/ic_vetkeys/src/encrypted_maps/canister.rs
```diff
@@ -127,6 +283,38 @@ macro_rules! export_encrypted_maps_canister {
...
+        /// Run `f` with a mutable reference to the initialized `EncryptedMaps`.
+        ///
+        /// Gives access to the raw value mutators; when wrapping them, keep any
+        /// linked side-state updated in the same endpoint call.
+        #[allow(dead_code)]
+        fn with_encrypted_maps_mut<__EmR>(
+            f: impl FnOnce(&mut __EmEncryptedMaps<__EmAccessRights>) -> __EmR,
+        ) -> __EmR {
+            ENCRYPTED_MAPS.with_borrow_mut(|encrypted_maps| f(encrypted_maps.as_mut().unwrap()))
+        }
```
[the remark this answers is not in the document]
marc0olo: Done together with `with_encrypted_maps` — now `pub(crate)` + `expect(...)`.

backend/rs/canisters/ic_vetkeys_encrypted_maps_custom_canister/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+fn insert_encrypted_value_custom(
+    map_owner: Principal,
+    map_name: ByteBuf,
+    map_key: ByteBuf,
+    value: EncryptedMapValue,
+) -> Result<Option<EncryptedMapValue>, String> {
+    with_encrypted_maps_mut(|encrypted_maps| {
+        encrypted_maps.insert_encrypted_value(
+            ic_cdk::api::msg_caller(),
+            (map_owner, to_blob(map_name)?),
```
mraszyk, line 63: Could you please add comments what the semantics of these values is supposed to be? Same below please.

backend/rs/ic_vetkeys/src/encrypted_maps/canister.rs
```diff
@@ -58,8 +76,97 @@
...
+/// # Bring your own value endpoints (`custom_value_endpoints`)
+///
+/// Pass the `custom_value_endpoints` marker to generate everything **except**
+/// the endpoints that read or write encrypted map values. This is for the
+/// "wrap-and-extend" pattern: a canister that stores something *alongside* each
+/// value — for example a backend-authoritative metadata record per entry — and
+/// therefore must expose its own `*_with_metadata` read/write endpoints and keep
+/// the two stores consistent in a single call.
+///
+/// ```ignore
```
mraszyk, line 89: this example is duplicated, maybe we can just keep it checked-in as code and drop this instance inserted as a comment?

backend/rs/ic_vetkeys/src/encrypted_maps/canister.rs
```diff
@@ -68,6 +175,59 @@ macro_rules! export_encrypted_maps_canister {
...
+    // Control-plane only: caller provides its own value read/write endpoints.
+    (
+        $domain_separator:expr,
+        [
+            $memory_domain_separator:expr,
+            $memory_access_control:expr,
+            $memory_shared_keys:expr,
+            $memory_encrypted_maps:expr $(,)?
+        ],
+        custom_value_endpoints $(,)?
```
mraszyk, line 184: what if there's a typo in the user's annotation "custom_value_endpoints" - will it fail to compile or silently ignore the annotation?

mraszyk (approved):
How about not injecting #[init]/#[post_upgrade], but exposing functions that could be invoked from #[init]/#[post_upgrade] so that people could define their own #[init]/#[post_upgrade]?

backend/rs/canisters/ic_vetkeys_encrypted_maps_custom_canister/src/lib.rs
marc0olo: Added — both custom endpoints now document the semantics of `map_owner`/`map_name`/`map_key`/`value` (owner+name identify the namespaced map, key is the entry within it, value is the client-side-encrypted blob), the return value, and that access is checked against the caller.

backend/rs/ic_vetkeys/src/encrypted_maps/canister.rs
marc0olo: Good point — the inline example is `ignore`d so it can drift. Trimmed it to a minimal invocation snippet and now point to the compile-tested `ic-vetkeys-encrypted-maps-custom-canister` reference canister (and `password_manager_with_metadata` for the full linked-metadata pattern). Kept a short snippet rather than dropping it entirely so the docs.rs macro page still has an inline orientation.

backend/rs/ic_vetkeys/src/encrypted_maps/canister.rs
marc0olo: Verified empirically: a typo fails to compile, it is not silently ignored. Changing the marker to `custom_value_endpoint` produces:

```
error: no rules expected `custom_value_endpoint`
       ^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
error[E0425]: cannot find function `with_encrypted_maps_mut` in this scope
error[E0425]: cannot find function `with_encrypted_maps` in this scope
```

The typo'd token matches neither arm (the custom arm matches the literal `custom_value_endpoints`; the full arm permits nothing after the memory array but a trailing comma), so the macro doesn't expand at all — hence the primary error plus the cascade from the un-generated accessors. There is no path where a typo silently falls through to the full canister.

marc0olo:
On the `#[init]`/`#[post_upgrade]` suggestion (@mraszyk): it's a fair composability point, so rather than decide unilaterally we'd like a couple more opinions before merging. One data point in the meantime: the motivating `password_manager_with_metadata` would get *simpler* under the current injected custom form — its `#[init]`/`#[post_upgrade]` only call `init_encrypted_maps(key_name)` and keep a `KEY_NAME` cell to feed the key back on upgrade, but the injected `#[post_upgrade]` already re-attaches by loading the key from EncryptedMaps' own config cell, and the metadata store self-initializes in its thread-local. So it would drop both hooks and the `KEY_NAME` cell entirely.

Exposing setup fns instead would help adopters who need custom init *args* or post-upgrade migration logic (none in the current use case), at the cost of more lifecycle boilerplate and reintroducing the 'forgot `#[post_upgrade]` → traps after upgrade' footgun we fixed in #421. Leaving it open for input; the other review comments are addressed.

adamspofford-dfinity (approved):

PR #425, opened 2026-08-04
feat(motoko): EncryptedMapsControlPlaneCanister mixin (wrap-and-extend)

marc0olo:
Adds Motoko canister mixins for EncryptedMaps, resolving the Motoko side of #423.

## Mixins
- **`EncryptedMapsCanister`** (`mo:ic-vetkeys/encrypted_maps/Canister`) — a complete EncryptedMaps canister. The adopter declares its own `EncryptedMapsState` stable variable and passes it in via `include EncryptedMapsCanister(encryptedMapsState)`, which adds every shared/query endpoint — so a canister's `Main.mo` is a few lines instead of ~200 lines of hand-written delegation. The mixin is the single source of the endpoint set, so the exposed Candid matches what the `@icp-sdk/vetkeys` frontend expects, by construction.
- **`EncryptedMapsControlPlaneCanister`** (`mo:ic-vetkeys/encrypted_maps/ControlPlaneCanister`) — given the caller's `EncryptedMapsState`, it provides the `encryptedMaps` instance, the `ByteBuf`/`Result` types, and the **8 control-plane endpoints** (vetKD keys, access control, map-name enumeration), but **none** of the 7 value read/write endpoints. The full `EncryptedMapsCanister` mixin is this mixin plus the value endpoints (composition — no duplication).

## Adopter owns the stable state
Neither mixin declares stable state of its own. The adopter declares the `EncryptedMapsState` in the actor body and passes it to the mixin; the mixin only builds the `transient` `encryptedMaps` wrapper over it and adds the endpoints. This keeps the persistent state a plain, visible stable variable the canister owns and can migrate — non-transient state buried inside a mixin is tricky under orthogonal persistence (thanks @alexandru-uta for the review). This mirrors the Rust macro, where the adopter already owns the persistent state (the `MemoryManager` + stable memories) and the macro only holds a transient handle.

```motoko
// Full canister:
import EncryptedMapsCanister "mo:ic-vetkeys/encrypted_maps/Canister";
import EncryptedMaps "mo:ic-vetkeys/encrypted_maps/EncryptedMaps";
import Types "mo:ic-vetkeys/Types";
import Runtime "mo:core/Runtime";

persistent actor {
    // `transient`: only an install-time input, baked into the stable state below.
    transient let keyName = Runtime.envVar<system>("VETKD_KEY_NAME") ?? "test_key_1";
    let encryptedMapsState = EncryptedMaps.newEncryptedMapsState<Types.AccessRights>(
        { curve = #bls12_381_g2; name = keyName },
        "my_app_domain_separator",
    );
    include EncryptedMapsCanister(encryptedMapsState);
};

// Wrap-and-extend: control plane + your own value endpoints, over the same state:
include EncryptedMapsControlPlaneCanister(encryptedMapsState);
// `encryptedMaps`, `ByteBuf`, `Result` are in scope.
public shared (msg) func insert_encrypted_value_with_metadata(...) : async Result<?ByteBuf, Text> {
    let r = encryptedMaps.insertEncryptedValue(msg.caller, (map_owner, map_name.inner), map_key.inner, value.inner);
    // ...maintain the linked side-state in the same call...
};
```

## Key configuration
Where the vetKD key name comes from is the adopter's choice — it is not baked into the library. The reference canisters read it at initialization from the **`VETKD_KEY_NAME`** canister environment variable, which the deployer sets via canister settings (choosing e.g. `test_key_1` or `key_1` without a rebuild), defaulting to `test_key_1` via `??` so init stays total (a missing env var can never leave the canister half-initialized). Because the key is resolved this way, the canister stays a plain `persistent actor` — no actor class or install argument — which is compatible with enhanced migration.

The key name is **immutable for the life of the canister's data**: it feeds vetKD key derivation, so it is captured once into the stable `EncryptedMapsState` at install and never re-read. Changing `VETKD_KEY_NAME` on a later upgrade is silently ignored (the setting and the key in use can diverge); only a `reinstall`, which drops all state, changes the key. Since `test_key_1` is a valid mainnet key too, a production deploy that forgets to set the variable silently runs on it — adopters who'd rather fail than run on the wrong key should assert the expected key instead of defaulting.

To make that immutability structural rather than by-convention, `KeyManagerState.vetKdKeyId` is now declared without `var` (immutable). It is only ever set by `newKeyManagerState` and read thereafter, so a reassignment — which would orphan every already-encrypted value — no longer type-checks. This gives the mixin the same effective guarantee the Rust macro gets from encapsulation (its key id sits behind a read-only handle); listed under Changed in the CHANGELOG.

## The control-plane split
The omitted value data-plane (7 endpoints): writes `insert_encrypted_value`, `remove_encrypted_value`, `remove_map_values`; reads `get_encrypted_value`, `get_encrypted_values_for_map`, `get_all_accessible_encrypted_values`, `get_all_accessible_encrypted_maps`. Only the **writes** can break a linked invariant; the **reads** are omitted for API-surface hygiene. The 8 control-plane endpoints never read or write map values, so they are always safe to include.

## Requirements
`moc` 1.13.0 (for mixin composition and `<system>` in mixin bodies) and `mo:core` 2.6.1. Recorded under the unreleased Motoko `0.6.0`.

## Reference canisters
- `ic-vetkeys-encrypted-maps-canister` (full mixin) and `ic-vetkeys-encrypted-maps-custom-canister` (control-plane mixin + custom value endpoints) — both plain `persistent actor`s that declare their own state. The custom one is compile-checked in `backend-motoko.yml`.
- The shared pocket-ic harness sets `VETKD_KEY_NAME` via `update_canister_settings` before install.

## Verification (moc 1.13.0 + mo:core 2.6.1)
- Full `EncryptedMapsCanister` interface is the standard one — shared harness **19/19** against the Motoko wasm, including the EOP upgrade test.
- Manager canister **14/14**, library `mops test` **2/2**, both reference canisters build.
- `mops format` and `cargo fmt`: clean.


backend/mo/ic_vetkeys/src/encrypted_maps/ControlPlaneCanister.mo
```diff
@@ -0,0 +1,160 @@
...
+    // The vetKD key name comes from the canister environment (set at deploy
+    // time via canister settings), not from an install argument — so the
+    // canister needs no actor class. Trap rather than default to any key.
+    let keyName : Text = switch (Runtime.envVar<system>("VETKD_KEY_NAME")) {
+        case (?name) { name };
+        case null {
+            Runtime.trap("the VETKD_KEY_NAME canister environment variable is not set");
+        };
+    };
+    let encryptedMapsState = EncryptedMaps.newEncryptedMapsState<Types.AccessRights>({ curve = #bls12_381_g2; name = keyName }, domainSeparator);
```
alexandru-uta: even though this works, I would not recommend doing it; let's declare this state variable in the main actor body and pass it as a parameter to the mixin if really needed. Non-transient state in mixins is very tricky :)

marc0olo: Done — adopted this. Both mixins now take the `EncryptedMapsState` as a parameter and declare no stable state of their own; the adopter declares it in the actor body and passes it in via `include EncryptedMapsCanister(encryptedMapsState)`. The mixin only builds the `transient` `encryptedMaps` wrapper over it. Updated both mixins, both reference canisters, and the docs. Thanks for the catch — agreed that non-transient state buried in a mixin is a footgun. (This also lines up with the Rust macro, where the adopter already owns the persistent state.)

backend/mo/canisters/ic_vetkeys_encrypted_maps_custom_canister/src/Main.mo
```diff
@@ -0,0 +1,71 @@
...
+import EncryptedMaps "mo:ic-vetkeys/encrypted_maps/EncryptedMaps";
+import Types "mo:ic-vetkeys/Types";
+import Principal "mo:core/Principal";
+import Text "mo:core/Text";
+import Runtime "mo:core/Runtime";
+
+persistent actor {
+  // The canister owns its stable state. The vetKD key name comes from the
+  // `VETKD_KEY_NAME` canister environment variable, so no actor class is needed.
+  let keyName = switch (Runtime.envVar<system>("VETKD_KEY_NAME")) {
```
alexandru-uta: this is not super safe: initializing stable vars in the actor body runs at canister init time. if the env var is not set and you end up trapping (the other branch), then the other vars end up not being initialized (or maybe even some other weird behavior if you roll back the init). 

I think we should not do this. Maybe add another method called `initialize()` where we do all this dance.
Another option is to set the default branch to serve the test case always;

A third option, if you don't expect `keyName` to change under your feet during runtime via the canister env var, maybe it's even simpler to do: `transient let keyName = "test_key_1"` or vice-versa.

alexandru-uta: or maybe even:
` transient let keyName = (Runtime.envVar<system>("VETKD_KEY_NAME")) ?? "test_key_1" ; `

alexandru-uta: and this way you'll be the first user of Kamil's null-coallesce feature.

marc0olo: Adopted the `?? "test_key_1"` form you suggested — thanks, and yes, apparently we're the first users of Kamil's null-coalesce 🙂 (compiles fine on moc 1.13.0). Both reference canisters now do:

```motoko
transient let keyName = Runtime.envVar<system>("VETKD_KEY_NAME") ?? "test_key_1";
```

On the init-safety point: a trap in a stable-var initializer is actually atomic — the whole `install_code` message reverts, so you never get a half-initialized canister, and on upgrade the stable var already exists so its initializer doesn't re-run. So the old trap only ever fired at first install, cleanly. That said, a total init that never traps is nicer and removes the deploy friction, so your suggestion is the better call regardless.

Two deliberate details:
- `keyName` is `transient` because it is only an install-time input: it is baked into the stable `EncryptedMapsState` and never read again. On upgrade the transient re-runs and re-reads the env var, but the value is unused (the stable state isn't rebuilt) — so changing `VETKD_KEY_NAME` on a later upgrade is silently ignored, and the key is effectively frozen at install. Only a `reinstall` (which drops state) changes it. Documented this for adopters.
- Following the same "the key can't change under you" theme, I also made `KeyManagerState.vetKdKeyId` immutable (`let` instead of `var`), so a reassignment no longer type-checks. That gives the mixin the same effective guarantee the Rust macro gets from encapsulation.

All green: encrypted_maps 19/19 (Rust + Motoko incl. the EOP upgrade test), manager 14/14, library tests 2/2.

schneiderstefan (approved):

